PR #1, opened 2018-09-22
[ADD] task to set ready to merge label

sbidoul:
Extracted from https://github.com/OCA/maintainer-tools/pull/346

TODO

- [ ] ~cron to run the task nightly on all repos~ (or maybe not: we would quickly explode the API rate limit)
- [x] webhook

### DIFF
--- a/src/oca_github_bot/tasks/tag_ready_to_merge.py
+++ b/src/oca_github_bot/tasks/tag_ready_to_merge.py
@@ -1,0 +1,91 @@
+# Copyright (c) ACSONE SA/NV 2018
+# Distributed under the MIT License (http://opensource.org/licenses/MIT).
+
+from collections import defaultdict
+from datetime import datetime, timedelta
+
+from .. import github
+from ..github import gh_call
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+APPROVALS_REQUIRED = 2
+MIN_PR_AGE = timedelta(days=5)
+LABEL_APPROVED = "approved"
+LABEL_READY_TO_MERGE = "ready to merge"
+
+
+@task()
+def tag_ready_to_merge(org, repo, pr, dry_run=False):
+    with github.repository(org, repo) as gh_repo:
+        gh_pr = gh_call(gh_repo.pull_request, pr)
+        if not gh_pr.mergeable:
+            # TODO does this exclude merged and closed pr's?
+            # TODO remove approved and ready to merge labels here?
+            _logger.info(f"{gh_pr.url} is not mergeable, exiting")
+            return
+        reviews = list(gh_call(gh_pr.reviews))
+        # obtain last review state for each reviewer
+        review_state_by_user = {}  # login: last review state
+        for review in reviews:
+            if review.state == "COMMENTED":
+                # ignore "commented" review, as they don't change
+                # the review status
+                continue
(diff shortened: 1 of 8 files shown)


src/oca_github_bot/tasks/tag_ready_to_merge.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .. import github
+from ..github import gh_call
+from ..queue import task, getLogger
+
+_logger = getLogger(__name__)
+
+
+APPROVALS_REQUIRED = 2
+MIN_PR_AGE = 5  # days
```
damdam-s: Maybe we could use something like that to avoid counting down on Saturdays and Sundays by default.
https://gist.github.com/JosefJezek/275b2c667698aabd6969

src/oca_github_bot/tasks/tag_ready_to_merge.py
```diff
@@ -0,0 +1,36 @@
...
+from datetime import datetime, timedelta
+
+from .. import github
+from ..github import gh_call
+from ..queue import task, getLogger
+
+_logger = getLogger(__name__)
+
+
+APPROVALS_REQUIRED = 2
```
damdam-s: Maybe this value of `2` can be a fallback for a customized value on each repository.
This would allow more flexiblity and this kind" of rule could be managed by the PSC of the repo they are responsible of.

src/oca_github_bot/tasks/tag_ready_to_merge.py
sbidoul: Yes, that would be a nice improvement. I'll try to make the basic version work first, then we can improve.

src/oca_github_bot/tasks/tag_ready_to_merge.py
sbidoul: The question would be where to configure these parameters per repo. In any case, these label are only advisory.

sbidoul:
I reworked the logic to make a cron unncessary and only react on PR events.

sbidoul:
@OCA/board this PR is deployed since Saturday.

sbidoul:
This has been running fine for a week, was discussed on the board and approved by @max3903 so merging.

sbidoul:
@pedrobaeza it is handled correctly: there is a [task](https://github.com/OCA/oca-github-bot/blob/959b02d3cda6594e26808002885cb2381d5732b8/src/oca_github_bot/tasks/tag_ready_to_merge.py#L25) running [hourly](https://github.com/OCA/oca-github-bot/blob/959b02d3cda6594e26808002885cb2381d5732b8/src/oca_github_bot/cron.py#L18-L22) to set the ready to merge tag.

So if all conditions are met, the ready to merge tag will appear 5 days after creation of the PR.

pedrobaeza:
Ah, great, I thought that you removed all crons and only acts on hooks.

src/oca_github_bot/tasks/tag_ready_to_merge.py
sbidoul: @damdam-s if you think it's important to not count week-ends in the 5 days, can you open an issue for discussion?

damdam-s: @sbidoul #17 added to discuss

PR #6, opened 2018-10-01
[ADD] retry github tasks if they fail because of the rate limit

hbrunn:
@sbidoul so this is me making myself familiar with celery and pytests. I didn't yet test this live, but will work on doing this tomorrow.


pedrobaeza:
Can you check tests?

sbidoul:
@hbrunn looks good. Great idea to use Celery Retry exception for that.

The failing test is due to the black autoformatter complaining.
 I suggest you follow the `pre-commit` [install procedure](https://github.com/OCA/oca-github-bot#development) so it is applied automatically when you commit.

PR #7, opened 2018-10-02
[FIX] make environment.sample work for docker out of the box

hbrunn:
this will make it simpler for people to spin up a testing instance I think


sbidoul:
Thanks!

PR #22, opened 2018-11-18
Clabot fix rebase login

gurneyalex:


gurneyalex:
This PR fixes issues such as the ones reported in https://github.com/OCA/maintainer-tools/issues/394

Example of PR affected by the problem:
https://github.com/OCA/account-analytic/pull/204

Same PR (almost), running after the fix: https://github.com/OCA/account-analytic/pull/205

pedrobaeza:
Thanks for the effort and for including tests.

Issue #19, opened 2018-11-16
Tests for tasks and webhooks


sbidoul:
Maybe it's easy to do with [vcrpy](https://pypi.org/project/vcrpy/) or something similar.

pedrobaeza:
Well, mocking things we lose possible GitHub API changes. On the other side, we hit the rate limit each time a test is done.

sbidoul:
Example in #31

PR #31, opened 2018-12-28
add CLA bot in the new framework

### LINKED ISSUE #12: CLA Bot
Migrate the CLA bot to here.

gurneyalex:
closes #12


src/oca_github_bot/webhooks/on_pr_update_check_cla.py
```diff
@@ -0,0 +1,384 @@
...
+
+_logger = logging.getLogger(__name__)
+RE_WEBLATE_COMMIT_MSG = r'''^Translated using Weblate.*
+Translate-URL: https://translation\.odoo-community\.org/'''
+
+
+@router.register("pull_request")
+async def on_pr_check_cla(event, gh, *args, **kwargs):
+    checker = CLACheck(event, gh)
+    return checker.check_cla()
```
sbidoul: @gurneyalex Thanks! 

Could you move the bulk of the CLA bot to the tasks directory and decorate it with `@task()`, and here call it with `delay()`. That's the easiest to queue it, and avoid the need to asyncify everything.

gurneyalex: thanks @sbidoul I'll do this

src/oca_github_bot/webhooks/on_pr_update_check_cla.py
```diff
@@ -0,0 +1,383 @@
...
+
+"""
+
+
+def _init_database(dbfile):
+    with sqlite3.connect(dbfile) as db:
+        db.executescript(SQL_DATABASE_CREATE)
+
+
+_init_database(config.CLABOT_CACHE)
```
gurneyalex: @sbidoul I'm not quite sure how to handle this initialization. Do you have a suggestion?

src/oca_github_bot/webhooks/on_pr_update_check_cla.py
gurneyalex: done, still needs testing (I have not run this yet...)

src/oca_github_bot/tasks/check_cla.py
```diff
@@ -0,0 +1,385 @@
...
+_logger = logging.getLogger(__name__)
+RE_WEBLATE_COMMIT_MSG = r"""^Translated using Weblate.*
+Translate-URL: https://translation\.odoo-community\.org/"""
+
+
+_logger = getLogger(__name__)
+
+
+@task()
+def check_cla(event, gh):
```
sbidoul: Instead of event, consider passing explicit arguments such as org, repo, pr ([example](https://github.com/OCA/oca-github-bot/blob/e73d881f309a142ded4ee56068433c7b0eb2cc7a/src/oca_github_bot/tasks/tag_approved.py#L18)).

This makes it easier to invoke the task from the command line.

Also, gh can be obtained by reconnecting using global config (the task is potentially running in a different process).

src/oca_github_bot/webhooks/on_pr_update_check_cla.py
sbidoul: [main](https://github.com/OCA/oca-github-bot/blob/master/src/oca_github_bot/__main__.py) could call various initializers.

Or run it at the beginning of the task if the database does not exist?

gurneyalex:
test comment from api

gurneyalex:
Quite a bit of refactoring done this morning:

* extract the relevant bits from the event when creating the task
* string formatting (still not sure about the logging calls handling)
* use the github3 module to access the github API rather than raw HTTP


Not quite sure how to add tests on this. There's some mention of betamax on https://github3py.readthedocs.io/en/master/contributing/testing.html to help creating Mocks. I've never used it, though...  @sbidoul do you have an opinion on this?

sbidoul:
@gurneyalex 

Wrt testing, I was also thinking (#19) about using vcrpy or betamax. I know @guewen and @lmignon have experience with vcrpy. Here is a discussion about betamax vs vcrpy: https://github.com/OCA/oca-github-bot/issues/19.

I was thinking of doing separate tests for the webhook (with a regular call to the webhook function and classic mock of the task - [example](https://github.com/OCA/oca-github-bot/blob/master/tests/test_on_pr_close_delete_branch.py)). And a vcrpy mock for the task.

Wrt f-strings and logging... yes they are evaluated regardless of log level. But they are so much easier to read, so I tend to use them for INFO level logging where the performance impact is not significant.

gurneyalex:
I ended up using vcr for the tests. 

AFAICT the code works fine (see https://github.com/gurneyalex/server-tools/pull/1 for a test PR I used to record the cassette). 

If this gets merged, ping me before deploying. We need to:

* stop and disable the current clabot (running on runbot1)
* get the sqlite database of the current clabot and use it in the new version in order to avoid spurious notifications

in due time:
* remove the current clabot hooks on the repository as these are going to be replaced with the org-level hook
* update the "new repository" process of the OCA to remove the old clabot hook creation

src/oca_github_bot/config.py
```diff
@@ -22,3 +22,40 @@
 SENTRY_DSN = os.environ.get("SENTRY_DSN")
 
 DRY_RUN = os.environ.get("DRY_RUN", "").lower() in ("1", "true", "yes")
+
+CLABOT_CACHE = os.environ.get("CLABOT_CACHE", "clabot_cache.db")
+
+
+cla_ko_message = """Hey @{pull_user},
```
sbidoul: Use uppercase variable name for consistency with the rest of the config file.

src/oca_github_bot/__main__.py
```diff
@@ -41,6 +42,8 @@
 def main():
     # configure logging
     logging.basicConfig(level=logging.DEBUG)
+    # make sure the CLA database is available
+    init_cla_database(config.CLABOT_CACHE)
```
sbidoul: Could you move this initialization to the beginning of the task?

src/oca_github_bot/tasks/check_cla.py
```diff
@@ -0,0 +1,15 @@
...
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+_logger = getLogger(__name__)
+
+
+@task()
+def check_cla(owner, repo, pull_user, pr_number, action):
```
sbidoul: can you add a `@switchable` decorator (see examples in other tasks) so it can be disabled by config.

requirements.txt
```diff
@@ -30,5 +30,6 @@ six==1.11.0
 tornado==5.1.1
 uritemplate==3.0.0
 urllib3==1.24.1
+vcrpy==2.0.1
```
sbidoul: this is now in `test-requirements.txt` via pytest-vcr.

src/oca_github_bot/__main__.py
gurneyalex: probably. But this looks like waste of time, as:

1. this will need checking for the file each time the task is invoked
2. probably this will need locking to prevent async invocations of the task from getting in a race condition over the existence of the file. 

Are you sure this is worth it?

sbidoul: Hm, yes. Then could you decorate it with `@switchable('clabot')` as well as the cla bot task? So they can be both disabled with the same configuration.

gurneyalex:
@sbidoul rebased

gurneyalex:
@sbidoul I still see surprising things in odoo_client_.py Would like to discuss this with you :)

dreispt:
CLA handling came up in a discussion as being an expensive item for the OCA budget.
Is this PR still a thing?

sbidoul:
@dreispt It's still TODO, but we maybe should think about migrating to a ready-made solution such as https://github.com/cla-assistant/cla-assistant that would also have the benefit to streamline the CLA signature process.

legalsylvain:
> @dreispt It's still TODO, but we maybe should think about migrating to a ready-made solution such as https://github.com/cla-assistant/cla-assistant that would also have the benefit to streamline the CLA signature process.

@sbidoul Is it possible for an OCA tool to be based on a tool that says _"the best run SAP"_ ? ;-)
#SorrySAP

ref : https://github.com/cla-assistant/cla-assistant#credits

sbidoul:
Closing this per https://github.com/OCA/oca-github-bot/issues/170#issuecomment-1090208307

PR #34, opened 2019-03-21
[FIX] bump_version: reset minor digits when updating major digits

mreficent:
The test change shows what this fix means.


mreficent:
ping: @sbidoul

Issue #20, opened 2018-11-16
Set "Needs review" label automatically when CIs goes green


pedrobaeza:
The title says all :wink:

sbidoul:
Unless it has WIP in the title

pedrobaeza:
Good point, yes

PR #37, opened 2019-05-04
[ADD] master: needs review label to green pull requests

Angelfentanez:
This PR adds a method that listens to the `check_suite` webhooks triggered when the CI tests finish in a PR and adds the `needs review` label if the tests were successful and the PR's title doesn't start with `WIP:` (case insensitive). 

If another commit is pushed to the PR and the tests fail, the label is removed.

closes https://github.com/OCA/oca-github-bot/issues/20

### DIFF
--- a/src/oca_github_bot/tasks/tag_needs_review.py
+++ b/src/oca_github_bot/tasks/tag_needs_review.py
@@ -1,0 +1,33 @@
+from ..github import gh_call, repository
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+LABEL_NEEDS_REVIEW = "needs review"
+LABEL_WIP = "work in progress"
+
+
+@task()
+def tag_needs_review(org, pr, repo, status, dry_run=False):
+    """On a successful execution of the CI tests, adds the `needs review`
+    label to the pull request if it doesn't have `wip:` at the
+    begining of the title (case insensitive). Removes the tag if the CI
+    fails.
+    """
+    with repository(org, repo) as gh_repo:
+        gh_pr = gh_call(gh_repo.pull_request, pr)
+        gh_issue = gh_call(gh_pr.issue)
+        labels = [label.name for label in gh_issue.labels()]
+        has_wip = (
+            gh_pr.title.lower().startswith(("wip:", "[wip]")) or LABEL_WIP in labels
+        )
+        if status == "success" and not has_wip:
+            if dry_run:
+                _logger.info(f"DRY-RUN add {LABEL_NEEDS_REVIEW} label")
+            else:
+                gh_call(gh_issue.add_labels, LABEL_NEEDS_REVIEW)
+            return
+        if dry_run:
+            _logger.info(f"DRY-RUN remove {LABEL_NEEDS_REVIEW} label")
+        else:
+            gh_call(gh_issue.remove_label, LABEL_NEEDS_REVIEW)
--- a/README.rst
+++ b/README.rst
@@ -52,6 +52,13 @@ On Pull Request review
 When there are two approvals, set the ``approved`` label.
 When the PR is at least 5 days old, set the ``ready to merge`` label.
 
+On Pull Request CI status
+-------------------------
+
+When the CI in a Pull Request goes green, set the ``needs review`` label.
+When a Pull Request with the ``needs review`` label is updated and the CI
+fails, remove the ``needs review`` label.
+
 Commands
 --------
 
@@ -170,6 +177,7 @@ Contributors
 * Michel Raich <miquel.raich@eficent.com>
 * Florian Kantelberg <florian.kantelberg@initos.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
+* Jose Angel Fentanez <joseangel@vauxoo.com>
 
 Maintainers
 ===========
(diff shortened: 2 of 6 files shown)


src/oca_github_bot/tasks/tag_needs_review.py
```diff
@@ -0,0 +1,31 @@
...
+    label to the pull request if it doesn't have `wip:` at the
+    begining of the title (case insensitive). Removes the tag if the CI
+    fails.
+    """
+    with repository(org, repo) as gh_repo:
+        gh_pr = gh_call(gh_repo.pull_request, pr)
+        gh_issue = gh_pr.issue()
+        if status == "success" and not gh_pr.title.lower().startswith("wip:"):
+            if dry_run:
+                _logger.info(f"DRY-RUN add {LABEL_NEEDS_REVIEW} label")
```
moylop260: Use code py3.5 friendly

src/oca_github_bot/tasks/tag_needs_review.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) ACSONE SA/NV 2018
```
moylop260: Are you sure?

Angelfentanez: LOL, used another file as base for this and forgot to remove this

src/oca_github_bot/tasks/tag_needs_review.py
Angelfentanez: @moylop260 I did it like this to keep it consistent with the other files:     
image

and also, the pre-commit config file uses python 3.6:    
image    

Should I change it to be python 3.5 compliant anyway?

moylop260: Ok
No problem
No changes required here then

moylop260:
could you check the red status of the build, please?
tests/test_on_pr_close_delete_branch.py .                                [ 73%]
tests/test_on_pr_green_label_needs_review.py 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Angelfentanez:
@moylop260  Done

sbidoul:
This looks good, thanks for this contribution. Can you update the README with a word about the new action, and add yourself in the contributors list?

sbidoul:
@Angelfentanez @moylop260 can you attend my last comment? Then we can merge.

Angelfentanez:
@sbidoul Sorry for the delay, changes done

sbidoul:
One other thing that occurred to me is the WIP tag could be checked too.
I'll merge and deploy after my Holidays.

src/oca_github_bot/tasks/tag_needs_review.py
```diff
@@ -0,0 +1,28 @@
...
+@task()
+def tag_needs_review(org, pr, repo, status, dry_run=False):
+    """On a successful execution of the CI tests, adds the `needs review`
+    label to the pull request if it doesn't have `wip:` at the
+    begining of the title (case insensitive). Removes the tag if the CI
+    fails.
+    """
+    with repository(org, repo) as gh_repo:
+        gh_pr = gh_call(gh_repo.pull_request, pr)
+        gh_issue = gh_pr.issue()
```
sbidoul:
```suggestion
        gh_issue = gh_call(gh_pr.issue)
```

src/oca_github_bot/tasks/tag_needs_review.py
```diff
@@ -0,0 +1,28 @@
...
+        if status == "success" and not gh_pr.title.lower().startswith("wip:"):
```
sbidoul: @Angelfentanez Could you also check for `"[wip]" in gh_pr.title.lower()` as it is frequently used?

Angelfentanez:
@moylop260 @sbidoul Sorry, had my hands full, suggested changes done.

src/oca_github_bot/tasks/tag_needs_review.py
```diff
@@ -0,0 +1,33 @@
...
+        if status == "success" and not has_wip:
+            if dry_run:
+                _logger.info(f"DRY-RUN add {LABEL_NEEDS_REVIEW} label")
+            else:
+                gh_call(gh_issue.add_labels, LABEL_NEEDS_REVIEW)
+            return
+        if dry_run:
+            _logger.info(f"DRY-RUN remove {LABEL_NEEDS_REVIEW} label")
+        else:
+            gh_call(gh_issue.remove_label, LABEL_NEEDS_REVIEW)
```
sbidoul: I'm not sure removing the needs review label is a good thing to do.
Indeed, the label may have been added manually.

sbidoul:
Following work on the merge bot, it became apparent that [checking for a green status is not as obvious as it seems](https://github.com/OCA/oca-github-bot/blob/74aae15da41a06b1969e30d3410a9bb66326e3a8/src/oca_github_bot/tasks/merge_bot.py#L160-L197) (multiple check suites, mix of new style checks and old style statuses, both travis versions being active at the same time...).

So maybe we should factor out the method that checks statuses in a separate tasks that dispatches to the merge bot (on red, on green) and this need-review action (on green).

sbidoul:
> So maybe we should factor out the method that checks statuses

Or maybe not, check_suite may be enough on PR's. I merged manually after removing the part the removes the needs review label. Let's see how it works.

If you think it's useful to remove the needs review label, can you please open a new issue to discuss the spec?

Thanks for contributing to this project!

Issue #24, opened 2018-11-30
Merge on demand


sbidoul:
A new bot task, that would be triggered by a command in a pull request comment.

**Command syntax**

`/ocabot merge (major|minor|patch)?`

**Actions performed**

- on merge command detected in PR comment
  - !!! check the author of the comment has the right to merge
    - is part of a team with write permission on the repo
    - or, is declared as maintainer in the manifest of all addons modified by the PR
  - rebase on main branch with autosquash
  - generate addon readme and setup.py
  - bump manifest version if (major|minor|patch), looking at all addons changed by the PR
  - push to a new branch named ``ocabot-merge-pr<pr>-to-<branch>`` (fail if that branch already exists)

- on red CI status on ``ocabot-merge-pr<pr>-to-<branch>`` branch
  - (ignore coverage/coveralls) 
  - add failed comment on PR
  - delete branch, exit

- on green CI status on ``ocabot-merge-pr<pr>-to-<branch>`` branch
  - push to main branch (--no-ff), if fail: rebase on main branch and force push ``ocabot-merge-pr<pr>-to-<branch>``, if rebase fails, delete branch and exit
  - delete branch ``ocabot-merge-pr<pr>-to-<branch>``
  - add a comment on the PR: *merged at sha*
  - close the PR
  - create wheel and publish to pypi, for the commit of our previous merge (to make sure a non .dev version goes to pypi if we bumped version)

**Attention point**
- makepot must somehow run before bumping the version: idea travis does makepot on `ocabot-merge-pr#`, and a local `msgmerge -U` is done before building the wheel so the translations in the wheel are aligned with the .pot.
- ~maybe we should allow makepot to run and push to PR branches~ => risk of merge conflict

**Tasks**

- [x] create main PR for this issue to which other subtasks can be contributed: #39
- [x] main method @sbidoul 
- [x] tie main method to command @sbidoul 
- [x] [function to bump manifest version](https://github.com/OCA/oca-github-bot/blob/cfe057f17a56a3a2a8827e11dda546546b780712/src/oca_github_bot/manifest.py#L65)
- [x] function to check if a github login has write access to the repo @fkantelberg 
- [x] webhook @fkantelberg 
- [x] function to detect addons modified by a PR, [start here](https://github.com/OCA/oca-github-bot/blob/cfe057f17a56a3a2a8827e11dda546546b780712/src/oca_github_bot/manifest.py#L93)
[...]

PR #42, opened 2019-06-15
merge bot improvements

sbidoul:
Follow-on for #24 

- [x] handle check suite and status events, which are used by different travis versions
- [x] bump github3.py version to have access to the check_suite api
- [x] add a merge label
- [x] on repos which use the travis-ci.org infrastructure while the travis-ci.com is configured at org level, `check_suites` return `queued` status forever while `statuses` returns `success`... not sure what to do with that yet
- [ ] improve error handing


setup.py
```diff
@@ -25,7 +25,7 @@
         "gidgethub",
         "appdirs",
         # GitHub client
-        "github3.py",
+        "github3.py>=1.3.0",
```
pedrobaeza: Why greater on equal here and pin in requirements.txt?

src/oca_github_bot/config.py
```diff
@@ -53,3 +53,6 @@ def func_wrapper(*args, **kwargs):
 #  delete_branch,tag_approved,tag_ready_to_merge,gen_addons_table,gen_addons_readme,
 #  gen_addons_icon,setuptools_odoo,merge_bot
 BOT_TASKS = os.environ.get("BOT_TASKS", "all").split(",")
+
+GITHUB_STATUS_IGNORED = ["ci/runbot", "codecov/project", "codecov/patch"]
```
pedrobaeza: I wouldn't ignore ci/runbot, as it can be red due to legit reasons.

src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -96,18 +132,7 @@ def merge_bot_status(org, repo, merge_bot_branch, sha):
...
-                if status.context in ("ci/runbot", "codecov/project", "codecov/patch"):
-                    # ignore
-                    continue
-                if status.state == "success":
-                    success = True
-                else:
-                    success = False
-                    break
+            success = _get_commit_success(gh_commit)
             if success is True:
```
pedrobaeza: Isn't enough?
```suggestion
            if success:
```

src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -76,11 +87,36 @@ def merge_bot_start(org, repo, pr, username, bumpversion=None, dry_run=False):
...
+    gh_check_suites = github.gh_call(gh_commit.check_suites)
+    for check_suite in gh_check_suites:
+        if check_suite.app.name in GITHUB_CHECK_SUITES_IGNORED:
+            # ignore
+            continue
+        if check_suite.conclusion == "success":
+            success = True
+        elif not check_suite.conclusion:
+            # not complete
+            return None
```
pedrobaeza: Is the check repeated later?

src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -96,18 +132,7 @@ def merge_bot_status(org, repo, merge_bot_branch, sha):
...
-                    continue
-                if status.state == "success":
-                    success = True
-                else:
-                    success = False
-                    break
+            success = _get_commit_success(gh_commit)
             if success is True:
                 _merge_bot_merge_pr(org, repo, merge_bot_branch)
             elif success is False:
```
pedrobaeza: You can indicate here as comment for later references that this specific check is needed for non finished tests to not be reported as failed.

setup.py
sbidoul: Because I want to express the minimum required version (the one that has the check suite api).
requirements.txt pins the known good versions.

src/oca_github_bot/tasks/merge_bot.py
sbidoul: no, because success is None means it's still in progress

src/oca_github_bot/config.py
sbidoul: I'm not sure, runbot tend to be red for technical issues more often than not these days. 
The person requesting the merge is supposed to check everything is okay first. This additional test run is only to check issues after rebase so I think it's better to have it relatively lightweight and as reliable as possible.

sbidoul:
@pedrobaeza regarding the merged status on the PR, I think the best is to add a `merged` label. That's what they do at Odoo, it should be fine for OCA too.

src/oca_github_bot/tasks/merge_bot.py
pedrobaeza: Yeah, I know, but for truly expression (`None` is evaled as falsy), it doesn't matter.

pedrobaeza:
Sorry, but I don't like that. It's very uncomfortable to have all the PRs as closed (even with the tag), without traceability at commit level, and so on. Have you tried the following:

- Rebase the branch.
- Do merge operation with merge commit.
- Push to target branch.

to see if the pull request is marked as merged? With this, we will have the same flow (not pushing to other branches), but preserving GitHub statuses.

sbidoul:
The 3 steps you mention are exactly what the bot does: there is a merge commit and a link to the PR.
If the rebase is a fast forward, then github will mark it as merged, if not a fast forward github we close the PR. There is no better solution I know of.

We have have good visibility and queryability with a label, and full traceability of the commits that have been merged and a link to the PR thanks to the merge commit: see example  https://github.com/OCA/mis-builder/pulls?q=is%3Apr+is%3Aclosed and https://github.com/OCA/mis-builder/pull/195#issuecomment-502361980

sbidoul:
@pedrobaeza have you investigated if we can migrate all repos to the new travis-ci.com infrastructure?
The information I get from the github api seems to be unreliable in a repo such as mis-builder which uses gitlab-ci.org while gitlab-ci.com is enabled at the OCA organization level.

pedrobaeza:
> If the rebase is a fast forward, then github will mark it as merged, if not a fast forward github we close the PR. 

And what prevent to perform the rebase as fast forward?

> have you investigated if we can migrate all repos to the new travis-ci.com infrastructure?

No, I will and let you know.

sbidoul:
> And what prevent to perform the rebase as fast forward?

For instance if there are new commits in the main branch since the PR was created.

pedrobaeza:
But the only possible problem is a red branch in that case, isn't it? I don't see that as critical while we have proper status in PRs. Or isn't any way in GitHub API to mark a PR as merged manually?

sbidoul:
Another reason why we need the rebase is bump version.

> Or isn't any way in GitHub API to mark a PR as merged manually?

Not that I know of.

pedrobaeza:
I have been trying about migrating all to travis-ci.com, but migration is not still enabled in our repo. You can see instructions in:

https://docs.travis-ci.com/user/migrate/open-source-repository-migration#the-migration-steps

but going to https://travis-ci.com/organizations/OCA/repositories, there's no "Migrate" tab. Only new repositories are listed there.

sbidoul:
> But the only possible problem is a red branch in that case, isn't it?

I think preventing the main branches becoming red when merging PRs that have been open for a long time is one of the key benefits of the merge bot.

This version is deployed now. Here is an example of what it did: https://github.com/OCA/mis-builder/pull/196#issuecomment-502401110 . Note the merge commit is red, this is "normal" and comes from makepot. I'll modify the travis config so makepot runs on merge bot branches too, so we'll not have these errors anymore and that will be another benefit of the merge bot.

Some careful tests are welcome. If things don't work as expected, please notify me with the exact time of the merge command, so I can investigate in the log and webhook deliveries.

sbidoul:
@pedrobaeza thanks for investigating travis. I think I could implement a [workaround](https://github.com/OCA/oca-github-bot/pull/42/commits/bcb06c21c0abaa18aa7ec6d8eac0d8bdcf5341f2).

pedrobaeza:
Yes, please integrate setup, POT, and translations update in the same rebase operation for at least having that advantage, or we will be wasting a lot of Travis builds.

src/oca_github_bot/tasks/merge_bot.py
pedrobaeza: I mean with this if the bot repeats the check when CIs finish

pedrobaeza:
> thanks for investigating travis. I think I could implement a workaround.

Well, if you have experimented for seeing it properly implemented, let's go ahead and see possible problems on use.

One question: who can put the merge command on the PR? Only members with write access on the repository?

sbidoul:
> One question: who can put the merge command on the PR? Only members with write access on the repository?

[You need push permission on the repo otherwise your command is rejected](https://github.com/OCA/oca-github-bot/blob/bcb06c21c0abaa18aa7ec6d8eac0d8bdcf5341f2/src/oca_github_bot/tasks/merge_bot.py#L71-L77).

pedrobaeza:
Great, that's the expected one!

pedrobaeza:
Can we instruct the bot to squash on merge?

sbidoul:
@pedrobaeza if the fixups are correct in the commits they will be autosquashed.

To force a squash of all commits we need some way to tell it the commit message to use. Some inspiration could be drawn from the Odoo SA bot for that I suppose. Could you open another issue to draft and discuss the specification?

pedrobaeza:
Done in #43

pedrobaeza:
Should we move forward with this PR? Some commit squashing?

sbidoul:
This branch should be good enough, let's merge it.

/ocabot merge

pedrobaeza:
Merge bot merging its improvements. This is merge-inception! xd

src/oca_github_bot/config.py
pedrobaeza: @sbidoul I still think we should add this check, as now the merge command serves as "merge this when you finish", and we don't wait for tests results, and maybe a runbot warning can find its way to the main branch.

sbidoul:
Maybe. BTW could we make Travis red on the same warnings? I mean, runbot is
for manual tests, and Travis is for CI, right? So green Travis should be
sufficient.

Le mer. 26 juin 2019 à 19:44, Pedro M. Baeza <notifications@github.com> a
écrit :

pedrobaeza:
Not exactly, as that checks requires an Odoo running as they are impossible to be handled statically.

sbidoul:
I don't understand. Travis is testing in a running Odoo. Do you have an
example of such check that runbot does and not Travis?

Le mer. 26 juin 2019 à 20:03, Pedro M. Baeza <notifications@github.com> a
écrit :

pedrobaeza:
For example the warning about missing description in a model. The problem is that we can't only parse Travis log, and searching for "WARNING:" lines will provoke incorrect red status. Runbot warning logs, however, are properly detected by runbot and put in that state.

PR #56, opened 2019-07-07
merge bot: improve abort message

sbidoul:


src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -221,8 +221,12 @@ def merge_bot_status(org, repo, merge_bot_branch, sha):
                 gh_pr = gh.pull_request(org, repo, pr)
                 github.gh_call(
                     gh_pr.create_comment,
-                    f"@{username} Merge command aborted due to a failed check on "
-                    f"[{merge_bot_branch}]"
-                    f"(https://github.com/{org}/{repo}/commits/{sha}).",
+                    f"@{username} your merge command was aborted due to failed "
+                    f"check(s), which you can inspect on "
```
pedrobaeza: Maybe you can add `... check(s) on rebased branch over current main one` or similar

sbidoul: I think this would make the text heavier, without much added value, because the name of the branch is in the link, and the mention of the rebase is alway in the previous comment anyway.

pedrobaeza: OK then

sbidoul:
deployed

Issue #53, opened 2019-07-01
merge bot: parse command inside a greater message


pedrobaeza:
For now, the merge bot only reacts to the merge command putting that comment in an isolated message (example of greater message: https://github.com/OCA/account-invoicing/pull/552#issuecomment-507203903).

sbidoul:
#49 is tracking the handling of approval comments and can be handled independently.

pedrobaeza:
Right, I didn't remember I already put that one. Renaming the issue.

PR #61, opened 2019-08-03
Allow addon maintainers to merge

sbidoul:
* better detection of modified addons
* allow addons maintainer to request merge (#53)


src/oca_github_bot/manifest.py
```diff
@@ -85,6 +87,18 @@ def set_manifest_version(addon_dir, version):
...
+def is_maintainer(username, addon_dirs):
+    for addon_dir in addon_dirs:
+        try:
+            manifest = get_manifest(addon_dir)
+        except NoManifestFound:
+            return False
+        maintainers = manifest.get("maintainers", [])
+        if username not in maintainers:
+            return False
+    return True
```
max3903: @sbidoul Does this mean that if you are the maintainer of a module in a repo, you can merge any module of that repo?

sbidoul: no, only pr that only touch addons where you are maintainer in the target branch

sbidoul: see `_user_can_merge()`

Issue #10, opened 2018-11-06
Auto close old inactive pull requests


sbidoul:
PR without activity for more than 6 months can be closed automatically with a friendly comment.

legalsylvain:
@sbidoul could you close this one ?

PR #67, opened 2019-08-27
[IMP] Close stale PRs

alexey-pelykh:
Attempt to address #10

### DIFF
--- a/src/oca_github_bot/tasks/close_stale.py
+++ b/src/oca_github_bot/tasks/close_stale.py
@@ -1,0 +1,48 @@
+# Copyright 2019 Brainbean Apps (https://brainbeanapps.com)
+# Distributed under the MIT License (http://opensource.org/licenses/MIT).
+
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from .. import github
+from ..config import switchable
+from ..github import gh_call, gh_datetime
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+INACTIVITY_PERIOD = relativedelta(years=1)
+LABEL_STALE = "stale"
+CLOSING_AS_STALE_COMMENT = (
+    "This PR had no activity for a long time. 😴 "
+    "It's therefore being closed as stale. 💔 "
+    "Please, reopen if this PR is still relevant. 🤖"
+)
+
+
+@task()
+@switchable()
+def close_stale(org, repo=None, dry_run=False):
+    """Close all stale PRs where conditions are met."""
+    with github.login() as gh:
+        query = [
+            "type:pr",
+            "state:open",
+            f"updated:<{gh_datetime(datetime.utcnow() - INACTIVITY_PERIOD)}",
+        ]
+        if repo:
+            query.append(f"repo:{org}/{repo}")
+        else:
+            query.append(f"org:{org}")
+        for issue in gh.search_issues(" ".join(query)):
+            if dry_run:
+                _logger.info(f"DRY-RUN close PR {issue.html_url} as stale")
+            else:
+                _logger.info(f"close PR {issue.html_url} as stale")
+                gh_issue = issue.issue
+                gh_call(gh_issue.add_labels, LABEL_STALE)
+                gh_pr = gh_issue.pull_request()
+                gh_call(gh_pr.create_comment, CLOSING_AS_STALE_COMMENT)
+                gh_call(gh_pr.close)
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -7,4 +7,4 @@ combine_as_imports=True
 use_parentheses=True
 line_length=88
 known_first_party=oca_github_bot
-known_third_party = aiohttp,appdirs,celery,gidgethub,github3,odoorpc,pytest,requests,setuptools
+known_third_party = aiohttp,appdirs,celery,dateutil,gidgethub,github3,odoorpc,pytest,requests,setuptools
--- a/README.rst
+++ b/README.rst
@@ -61,6 +61,11 @@ On Pull Request CI status
 When the CI in a Pull Request goes green, set the ``needs review`` label,
 unless it has ``wip:``  or ``[wip]`` in it's title.
 
+Close stale PR
+--------------
+
+If PR had no activity for more than a year, it's closed as stale.
+
 Commands
 --------
 
@@ -191,6 +196,7 @@ Contributors
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Jose Angel Fentanez <joseangel@vauxoo.com>
 * Simone Rubino <simone.rubino@agilebg.com>
+* Alexey Pelykh <alexey.pelykh@brainbeanapps.com>
 
 Maintainers
 ===========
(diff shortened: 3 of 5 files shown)


alexey-pelykh:
@sbidoul any hint what's wrong? I've used the isort to apply the rules, it seems

sbidoul:
That was quick :)

The easiest is to run `pre-commit install` at the root of your local clone so the pre-commit runs when you commit and before pushing.

You can also run `pre-commit run --all-files` and commit the changes.

src/oca_github_bot/tasks/close_stale.py
```diff
@@ -0,0 +1,46 @@
...
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+INACTIVITY_PERIOD = relativedelta(years=1)
+LABEL_STALE = "stale 🗃"
+CLOSING_AS_STALE_COMMENT = (
+    "This PR had no activity for a long time. "
+    "It's therefore being closed as stale. 🤖"
```
lreficent: Wouldn't be suitable to also add something like "Please, reopen if this PR is still relevant"?

alexey-pelykh: :) That's so positive ))

lreficent: Better to be! People sometimes gets frustrated because it takes a lot of time to get something merged in OCA if you don't have reviews :sweat_smile:

alexey-pelykh: Yeah, politics

alexey-pelykh: Politics 🤷‍♂

alexey-pelykh:
Any idea how to make a test for this?

sbidoul:
I started doing some tests with `vcr`. There is an example that is merged, and another in #31.

It's actually easier than it sounds. You write a test with call to the task on a real repo with vcr in record mode. It will record the http calls and that's it.

Not ultra-elegant and such tests tend to be a bit fragile and test too much the code structure over the end result. But it does the job and it's better than no test. Alternatively, some heavy mocking can do the job too.

alexey-pelykh:
@sbidoul cool, but that would mean actually closing old PRs on some repo :) Just to ensure I got that right - execute actual (not dry-run or anything) on real OCA repo and record that?
Is there a quick how-to on making that recording?

sbidoul:
> but that would mean actually closing old PRs on some repo

You can also run it on a sandbox repo you own.

> Is there a quick how-to on making that recording?

The simplest howto is in the pytest-vcr readme: https://pypi.org/project/pytest-vcr/
That's pretty much all you need to record your first test.

jgrandguillaume:
Thanks @alexey-pelykh for contributing this feature !

src/oca_github_bot/tasks/close_stale.py
```diff
@@ -0,0 +1,47 @@
...
+            query.append(f"org:{org}")
+        for issue in gh.search_issues(" ".join(query)):
+            if dry_run:
+                _logger.info(f"DRY-RUN close PR {issue.html_url} as stale")
+            else:
+                _logger.info(f"close PR {issue.html_url} as stale")
+                gh_issue = issue.issue
+                gh_call(gh_issue.add_labels, LABEL_STALE)
+                gh_pr = gh_issue.pull_request()
+                gh_call(gh_pr.close, CLOSING_AS_STALE_COMMENT)
```
simahawk: what about posting a nice message saying "Closed because...blabla.... Feel free to reopen if you get back to it..." or something similar?

alexey-pelykh: Done

src/oca_github_bot/tasks/close_stale.py
```diff
@@ -0,0 +1,48 @@
...
+from ..github import gh_call, gh_datetime
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+INACTIVITY_PERIOD = relativedelta(years=1)
+LABEL_STALE = "stale 🗃"
+CLOSING_AS_STALE_COMMENT = (
+    "This PR had no activity for a long time. 😴"
```
Yajo:
```suggestion
    "This PR had no activity for a long time. 😴 "
```

src/oca_github_bot/tasks/close_stale.py
```diff
@@ -0,0 +1,48 @@
...
+    "It's therefore being closed as stale. 💔"
```
Yajo:
```suggestion
    "It's therefore being closed as stale. 💔 "
```

legalsylvain:
Hi @alexey-pelykh. Could you fix the conflict ?

alexey-pelykh:
@legalsylvain done

alexey-pelykh:
Related https://github.com/OCA/maintainer-tools/pull/444

Yajo:
I wonder... is it worth it to do our own bot instead of leveraging a GH action such as https://github.com/marketplace/actions/close-stale-issues? :thinking: 

I know the bot exists for its reasons but you know that story about wheels and inventions... :man_shrugging:

alexey-pelykh:
As long as it works and cleans the mess up, I don't care how 👍

sbidoul:
Using a bot maintained by GitHub is appealing and indeed the stale bot @Yajo  is pointing out seems to do the job. I like the two steps approach (warn first then close)

Two potential issues come to mind:

- maintaining the configuration across our hundred repos: for that we need to do it now with linter configs so we'll need to finish #85
- with crons I'm a bit concerned we'll hit the GitHub rate limit (5000 requests per hours): at some points we may want to distribute such crons over the day / week

alexey-pelykh:
Let's do something :) I generally agree that reinventing the wheel if there's a plugin already is dumb, yet there's one possible drawback: in case we want more complex logic of stale PRs, like "pre-mark those, and wait for another 6 months" or "alert all previous contributors" and stuff like that.

sbidoul:
Functionally the thing that bothers me with this PR is that there is no warning, and when the PR is closed the author cannot reopen it. With the stale bot there are two steps, and in between the author can comment to remove the stale label and prevent automatic closing.

alexey-pelykh:
Well, we can always refine this PR and add one more label, "possibly stale", and if there's a PR with this label and no activity for another 6 months, well, then it is. Though that would be making own stale bot

sbidoul:
Let's try the stale action by GitHub: https://github.com/OCA/mis-builder/commit/ea09b4c56e0312d3da839623d0ad0b631344aa50

sbidoul:
The [GitHub stale action](https://github.com/OCA/mis-builder/blob/10.0/.github/workflows/stale.yml) seems to be [working](https://github.com/OCA/mis-builder/pull/158#issuecomment-605626093).

alexey-pelykh:
Ok, then safe to close this PR and kill the branch?

Yajo:
I think yes. Let's continue in https://github.com/OCA/oca-addons-repo-template/pull/63.

sbidoul:
Agreed.

PR #70, opened 2019-09-03
merge bot: change merge strategy to "merge"

### LINKED ISSUE #68: merge strategies
Currently, the default and only available merge strategy for the merge command is `rebase`.

After further reflection, I now think the default merge strategy could be `merge`.
We can create the merge bot branch by branching off the target branch and merging the PR into it. The rest of the process can be identical and we keep the same benefits (i.e. testing on top of the main branch before merging into it, bumping version as last operation, etc). 

Additional benefits of the merge strategy would be that branches which contain merge conflict resolution should be mergeable via the bot command, and GitHub would see PRs as merged.

Other merge stragegies (rebase with autosquash, squash all) can be made available via additional options to the ocabot merge command.

sbidoul:
Closes #68 

Instead of rebasing the PR onto the target branch, create a merge commit to the main branch, and push it to the main branch if green.

So the first phase
- creates the merge bot branch from the target branch
- creates the merge commit of the PR to the target branch
- pushes the merge bot branch
- await test results

And the second phase
- checks if the target branch is an ancestor of the merge bot branch (else retry the first phase)
- bump version
- push the merge bot branch to the target branch

### DIFF
--- a/src/oca_github_bot/tasks/merge_bot.py
+++ b/src/oca_github_bot/tasks/merge_bot.py
@@ -1,8 +1,10 @@
 # Copyright (c) ACSONE SA/NV 2019
 # Distributed under the MIT License (http://opensource.org/licenses/MIT).
 
+import os
 import random
 import subprocess
+from enum import Enum
 
 from .. import github
 from ..build_wheels import build_and_check_wheel, build_and_publish_wheel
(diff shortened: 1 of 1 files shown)


sbidoul:
I deployed this. Let's see.

sbidoul:
This seems to work well today, so merging.

PR #71, opened 2019-09-04
[ADD] rebase_bot

MiquelRForgeFlow:
New command. I hope it will be useful.

### DIFF
--- a/src/oca_github_bot/config.py
+++ b/src/oca_github_bot/config.py
@@ -109,7 +109,9 @@ if os.environ.get("OCABOT_TWINE_REPOSITORIES"):
 
 OCABOT_USAGE = os.environ.get(
     "OCABOT_USAGE",
-    "**Ocabot commands**\n" "* ``ocabot merge major|minor|patch|nobump``",
+    "**Ocabot commands**\n"
+    "* ``ocabot merge major|minor|patch|nobump``\n"
+    "* ``ocabot rebase``",
 )
 
 OCABOT_EXTRA_DOCUMENTATION = os.environ.get(
--- a/src/oca_github_bot/manifest.py
+++ b/src/oca_github_bot/manifest.py
@@ -5,7 +5,7 @@ import ast
 import os
 import re
 
-from .github import git_get_current_branch
+from .github import git_get_current_branch, github_user_can_push
 from .process import check_call, check_output
 
 MANIFEST_NAMES = ("__manifest__.py", "__openerp__.py", "__terp__.py")
@@ -210,3 +210,30 @@ def get_odoo_series_from_branch(branch):
     if not series:
         raise OdooSeriesNotDetected()
     return tuple(int(s) for s in series.split("."))
+
+
+def user_can_push(gh, org, repo, username, addons_dir, target_branch):
+    """
+    Check if the user is maintainer of all modified addons.
+
+    Assuming, addons_dir is a git clone of an addons repository,
+    return true if username is declared in the maintainers key
+    on the target branch, for all addons modified in the current branch
+    compared to the target_branch.
+    """
+    gh_repo = gh.repository(org, repo)
+    if github_user_can_push(gh_repo, username):
+        return True
+    modified_addon_dirs, other_changes = git_modified_addon_dirs(
+        addons_dir, target_branch
+    )
+    if other_changes or not modified_addon_dirs:
+        return False
+    # if we are modifying addons only, then the user must be maintainer of
+    # all of them on the target branch
+    current_branch = git_get_current_branch(cwd=addons_dir)
+    try:
+        check_call(["git", "checkout", target_branch], cwd=addons_dir)
+        return is_maintainer(username, modified_addon_dirs)
+    finally:
+        check_call(["git", "checkout", current_branch], cwd=addons_dir)
--- a/src/oca_github_bot/tasks/rebase_bot.py
+++ b/src/oca_github_bot/tasks/rebase_bot.py
@@ -1,0 +1,111 @@
+# Copyright (c) ForgeFlow, S.L. 2021
+# Distributed under the MIT License (http://opensource.org/licenses/MIT).
+
+from .. import config, github
+from ..config import switchable
+from ..manifest import user_can_push
+from ..process import CalledProcessError, check_call
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+@task()
+@switchable("rebase_bot")
+def rebase_bot_start(org, repo, pr, username, dry_run=False):
+    with github.login() as gh:
+        gh_pr = gh.pull_request(org, repo, pr)
+        target_branch = gh_pr.base.ref
+        remote = gh_pr.head._repo_owner
+        pr_branch = f"tmp-pr-{pr}"
+        try:
+            with github.temporary_clone(org, repo, target_branch) as clone_dir:
+                if not remote:
+                    github.gh_call(
+                        gh_pr.create_comment,
+                        f"Sorry @{username}, impossible to rebase because "
+                        f"unknown remote.",
+                    )
+                    return
+                check_call(
+                    ["git", "fetch", "origin", f"pull/{pr}/head:{pr_branch}"],
+                    cwd=clone_dir,
+                )
+                check_call(["git", "checkout", pr_branch], cwd=clone_dir)
+                if not user_can_push(gh, org, repo, username, clone_dir, target_branch):
+                    github.gh_call(
+                        gh_pr.create_comment,
+                        f"Sorry @{username} you are not allowed to rebase.\n\n"
+                        f"To do so you must either have push permissions on "
+                        f"the repository, or be a declared maintainer of all "
+                        f"modified addons.\n\n"
+                        f"If you wish to adopt an addon and become it's "
+                        f"[maintainer]"
+                        f"(https://odoo-community.org/page/maintainer-role), "
+                        f"open a pull request to add "
+                        f"your GitHub login to the `maintainers` key of its "
+                        f"manifest.",
+                    )
+                    return
+                # do rebase
+                check_call(["git", "fetch", "origin", target_branch], cwd=clone_dir)
+                check_call(["git", "rebase", f"origin/{target_branch}"], cwd=clone_dir)
+                # push rebase
+                true_pr_branch = gh_pr.head.ref
+                if dry_run:
+                    _logger.info(
+                        f"DRY-RUN git push in {remote}/{repo}@{true_pr_branch}"
+                    )
+                else:
+                    _logger.info(f"git push in {remote}/{repo}@{true_pr_branch}")
+                    check_call(
+                        [
+                            "git",
+                            "remote",
+                            "add",
+                            remote,
+                            f"https://github.com/{remote}/{repo}.git",
+                        ],
+                        cwd=clone_dir,
+                    )
+                    check_call(
+                        [
+                            "git",
+                            "remote",
+                            "set-url",
+                            "--push",
+                            remote,
+                            f"https://{config.GITHUB_TOKEN}@github.com/{remote}/{repo}",
+                        ],
+                        cwd=clone_dir,
+                    )
+                    check_call(
+                        [
+                            "git",
+                            "push",
+                            "--force",
+                            remote,
+                            f"{pr_branch}:{true_pr_branch}",
+                        ],
+                        cwd=clone_dir,
+                        log_error=False,
+                    )
+                github.gh_call(
+                    gh_pr.create_comment,
+                    f"Congratulations, PR rebased to [{target_branch}]"
+                    f"(https://github.com/{org}/{repo}/commits/{target_branch}).",
+                )
+        except CalledProcessError as e:
+            cmd = " ".join(e.cmd)
+            github.gh_call(
+                gh_pr.create_comment,
+                f"@{username} The rebase process failed, because "
+                f"command `{cmd}` failed with output:\n```\n{e.output}\n```",
+            )
+            raise
+        except Exception as e:
+            github.gh_call(
+                gh_pr.create_comment,
+                f"@{username} The rebase process failed, because of exception {e}.",
+            )
+            raise
--- a/README.rst
+++ b/README.rst
@@ -85,6 +85,10 @@ can be used to ask the bot to do the following:
 * when the version was bumped, generate a wheel, rsync it to a PEP 503
   simple index root, or upload it to one or more indexes with twine
 
+``/ocabot rebase`` can be used to ask the bot to do the following:
+
+* rebase the PR on the target branch
+
 TODO (help wanted)
 ------------------
 
@@ -205,7 +209,7 @@ Contributors
 
 * Stéphane Bidoul <stephane.bidoul@acsone.eu>
 * Holger Brunn <hbrunn@therp.nl>
-* Miquel Raïch <miquel.raich@eficent.com>
+* Miquel Raïch <miquel.raich@forgeflow.com>
 * Florian Kantelberg <florian.kantelberg@initos.com>
 * Laurent Mignon <laurent.mignon@acsone.eu>
 * Jose Angel Fentanez <joseangel@vauxoo.com>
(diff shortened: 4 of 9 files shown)


mreficent:
@sbidoul should I put manually the `bs4==4.8.0` in the requirements? if not, how should I solve the error?

mreficent:
@sbidoul ready to be reviewed! :)

requirements.txt
```diff
@@ -13,6 +13,7 @@ attrs==18.2.0             # via aiohttp
 babel==2.6.0              # via flower
 billiard==3.5.0.4         # via celery
 bleach==3.1.0             # via readme-renderer
+bs4==0.0.1
```
sbidoul: dependencies must go in setup.py first, and these frozen requirements are generated with `pip-compile` (from the pip-tools project) -- I should mention that in the README..

src/oca_github_bot/config.py
```diff
@@ -67,5 +68,11 @@ def func_wrapper(*args, **kwargs):
     "Hey, thanks for contributing! Proceeding to merge this for you.",
     "What a great day to merge this nice PR. Let's do it!",
 ]
+REBASE_BOT_INTRO_MESSAGES = [
+    "On my way to rebase this fine PR!",
+    "This PR looks fantastic, let's rebase it!",
+    "Hey, thanks for contributing! Proceeding to rebase this for you.",
+    "What a great day to rebase this nice PR. Let's do it!",
+]
```
sbidoul: Not sure we need this for a simple reabase :)

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,135 @@
...
+    if other_changes or not modified_addon_dirs:
+        return False
+    # if we are modifying addons only, then the user must be maintainer of
+    # all of them on the target branch
+    current_branch = github.git_get_current_branch(cwd=addons_dir)
+    try:
+        _git_call(["git", "checkout", target_branch], cwd=addons_dir)
+        return is_maintainer(username, modified_addon_dirs)
+    finally:
+        _git_call(["git", "checkout", current_branch], cwd=addons_dir)
```
sbidoul: I think this method and associated exception and comment on the PR should be factored out and generalized.
It's the same conditions for rebase or merge.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,135 @@
...
+            current_branch = github.git_get_current_branch(cwd=".")
+            if not intro_message:
+                intro_message = _get_rebase_bot_intro_message()
+            github.gh_call(
+                gh_pr.create_comment,
+                f"{intro_message}\n"
+                f"Rebasing to [{target_branch}]"
+                f"(https://github.com/{org}/{repo}/pull/{pr}/commits/{current_branch})"
+                f".",
+            )
```
sbidoul: Since everything is done in one step there is no need for an intro message. A success or error message is enough.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,135 @@
...
+            if bool(soup.find_all(name="span", attrs={"class": "unknown-repo"})):
+                github.gh_call(
+                    gh_pr.create_comment,
+                    f"Sorry @{username}, impossible to rebase "
+                    f"because unknown repository.",
+                )
+                return
+            span_list = soup.find_all(
+                name="span", attrs={"class": "css-truncate-target"}
+            )
```
sbidoul: Is there really no better way to find out the source branch and repo of the PR? That html parsing sounds very fragile.

src/oca_github_bot/config.py
mreficent: But it's fun, I like it.

src/oca_github_bot/tasks/rebase_bot.py
mreficent: No, because https://help.github.com/en/articles/checking-out-pull-requests-locally#error-failed-to-push-some-refs.

mreficent:
@sbidoul all comments attended.

I my opinion, I think it's not intrusive. This is just a tool. It depends of the maintainers on how they use the tool. If they are rebasing continuously around wantonly, then they may be in the intrusive side. But I think the correct use of this tool is to be used sporadically, when the creator of the PR is not rebasing or don't know how to rebase correctly. Or simply, the creator of the PR wants to rebase (but don't want to push anymore) so it call this tool :)

src/oca_github_bot/tasks/rebase_bot.py
sbidoul: The GitHub API gives you access to the source branch:

image

So you should be able to find the information through `gh.repository()`.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,94 @@
...
+                        "remote",
+                        "add",
+                        remote,
+                        f"https://github.com/{remote}/{repo}.git",
+                    ]
+                )
+                _git_call(["git", "push", "--force", remote, current_branch])
+                _git_call(["git", "remote", "remove", remote])
+            else:
+                _git_call(["git", "push", "--force", "origin", current_branch])
```
sbidoul: git push accepts a url, it should not be necessary to create a remote. Also no need for a different code path for the OCA remote, I think.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,94 @@
...
+                gh_pr.create_comment,
+                f"Congratulations, PR rebased to [{target_branch}]"
+                f"(https://github.com/{org}/{repo}/pull/{pr}/commits/{current_branch})"
+                f".",
+            )
+        except subprocess.CalledProcessError as e:
+            cmd = " ".join(e.cmd)
+            github.gh_call(
+                gh_pr.create_comment,
+                f"@{username} The rebase process could not start, because "
```
sbidoul: s/could not start/failed/

The "could not start" for the merge bot was because there was a second step (i.e. waiting for the checks).

src/oca_github_bot/tasks/rebase_bot.py
sbidoul: That said a remote is better because I think you will need to insert the github token in the url for this to work.

.isort.cfg
```diff
@@ -7,4 +7,4 @@ combine_as_imports=True
 use_parentheses=True
 line_length=88
 known_first_party=oca_github_bot
-known_third_party = aiohttp,appdirs,celery,gidgethub,github3,odoorpc,pytest,requests,setuptools
+known_third_party = aiohttp,appdirs,bs4,celery,gidgethub,github3,odoorpc,pytest,requests,setuptools
```
lreficent: I think is better to use the official package name (beautifulsoup4) instead of that dummy one. See https://pypi.org/project/bs4/. Don't you think?

mreficent: Ok, updated `setup.py` file and `requirements.txt` file accordingly. But this `isort.cfg` file I cannot change, because is filled automatically by `seed-isort-config` check, that interprets the `bs4` module like a package.

mreficent:
@sbidoul all comments attended :)

src/oca_github_bot/tasks/rebase_bot.py
mreficent: Oh good, I will try.

mreficent:
Why the clone? I don't understand. You fetch the branch, does the rebase on local and then push again to the same branch. It's a simple process, I don't see why adding more complicated steps. :S

mreficent:
Done now without using beautifulsoup 🎉🎉🎉🎉🎉🎉

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,90 @@
...
+            # fetch pr and create local branch
+            current_branch = github.git_get_current_branch(cwd=".")
+            _git_call(
+                [
+                    "git",
+                    "fetch",
+                    "origin",
+                    f"refs/pull/{pr}/head:{remote}_{current_branch}_{pr}",
+                ]
+            )
```
sbidoul: I may miss something obvious, but I don't see how this part can work unless cwd is a git repo. That's why I mentioned a needed call to temporary_clone before.

mreficent:
@sbidoul I have added the call to `temporary_clone`. What do you think now?

sbidoul:
It looks better. Travis is red though. Generally I'm concerned that you may not have done enough manual tests on this PR. I'll be around at #OCAdays to help you running the bot locally for testing.

mreficent:
> Travis is red though

Now is green :green_apple: 

> Generally I'm concerned that you may not have done enough manual tests on this PR

Yeah, right, I know. It's difficult to test it. The easy way is to implement and trying. 😁

> I'll be around at #OCAdays to help you running the bot locally for testing.

Good!!

sbidoul:
To facilitate review, could you extract the refactoring of `user_can_push` to a separate PR (or commit at least)?

mreficent:
@sbidoul done

MiquelRForgeFlow:
@sbidoul I have rebased. Please, remove the "needs rebase" tag and see if we can proceed with this.

Also, I see in the travis CI that something wrong happens with the python 3.7-dev, like as if it doesn't have pip. (not related to this PR)

pedrobaeza:
now s/python 3.7-dev/python 3.7 :wink

MiquelRForgeFlow:
thanks, now it's green :)

legalsylvain:
Hi @MiquelRForgeFlow. Could you fix the conflict ?

MiquelRForgeFlow:
rebased and fixed

src/oca_github_bot/manifest.py
```diff
@@ -199,3 +199,26 @@ def get_odoo_series_from_version(version):
     if not series:
         raise OdooSeriesNotDetected()
     return tuple(int(s) for s in series.split("."))
+
+
+def _user_can_push(gh, org, repo, username, addons_dir, target_branch):
```
legalsylvain: why this function begins with "_" ? the others function don't.

MiquelRForgeFlow: It's a rename of `_user_can_merge`. And why `_user_can_merge` begins with "_"? I don't know. But you can ask to @sbidoul

src/oca_github_bot/manifest.py
```diff
@@ -199,3 +199,26 @@ def get_odoo_series_from_version(version):
...
+    """
+    Check if a user is allowed to push.
+    addons_dir must be a git clone of the PR's branch.
```
sbidoul: Check if the user is maintainer of all modified addons.

Assuming, ``addons_dir`` is a git clone of an addons repository, return true if ``username`` is declared in the ``maintainers`` key on the target branch, for all addons modified in the current branch compared to the ``target_branch``.

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: python
 
 python:
   - "3.6"
-  - "3.7-dev"
+  - "3.7"
```
sbidoul: Please don't change this in this PR.

requirements.txt
```diff
@@ -50,4 +50,4 @@ wheel==0.33.4
 yarl==1.2.6               # via aiohttp
 
 # The following packages are considered to be unsafe in a requirements file:
-# setuptools==41.0.1
+# setuptools==41.2.0
```
sbidoul: This sounds unrelated, please don't change it in this PR.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,92 @@
...
+                local_branch = f"{remote}_{current_branch}_{pr}"
+                check_call(
+                    ["git", "fetch", "origin", f"refs/pull/{pr}/head:{local_branch}"],
+                    cwd=clone_dir,
+                )
+                check_call(["git", "checkout", current_branch], cwd=clone_dir)
+                # do rebase
+                check_call(["git", "fetch", "origin", target_branch], cwd=clone_dir)
+                check_call(
+                    ["git", "rebase", "--autosquash", target_branch], cwd=clone_dir
```
sbidoul: `--autosquash` only works in interactive mode. For for this to work, you need to set an environment variable `EDITOR=true` so a "do nothing" editor is launched and the actions applied automatically.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,92 @@
...
+@switchable("rebase_bot")
+def rebase_bot_start(org, repo, pr, username, dry_run=False):
+    with github.login() as gh:
+        gh_pr = gh.pull_request(org, repo, pr)
+        target_branch = gh_pr.base.ref
+        try:
+            with github.temporary_clone(org, repo, target_branch) as clone_dir:
+                if not _user_can_push(
+                    gh, org, repo, username, clone_dir, target_branch
+                ):
```
sbidoul: `user_can_push` is looking at modified addons in the current branch compared to `target_branch`. Here you have cloned `target_branch` there will no modified addons and `user_can_push` will always return false unless the user has write permissions on the repo.

So you need to checkout the PR branch.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,92 @@
...
+                remote = gh_pr.head._repo_owner
+                if not remote:
+                    github.gh_call(
+                        gh_pr.create_comment,
+                        f"Sorry @{username}, impossible to rebase "
+                        f"because unknown remote.",
+                    )
+                    return
+                # fetch pr and create local branch
+                current_branch = github.git_get_current_branch(cwd=clone_dir)
```
sbidoul: I'm not sure this is correct. If I read correctly, current_branch is going to be the same as target_branch, because that is what has been checked out in temporary_clone above, it has not changed since.

src/oca_github_bot/tasks/rebase_bot.py
```diff
@@ -0,0 +1,92 @@
...
+                        "git",
+                        "push",
+                        "--force",
+                        f"https://github.com/{remote}/{repo}.git",
+                        f"{local_branch}:{current_branch}",
+                    ],
+                    cwd=clone_dir,
+                    log_error=False,
+                )
+                check_call(["git", "branch", "-D", f"{local_branch}"], cwd=clone_dir)
```
sbidoul: Don't bother deleting the local branch, since we are in a temporary clone directory that will deleted automatically.

.travis.yml
MiquelRForgeFlow: but 3.7-dev gave error of not existing putting the PR in red :(

MiquelRForgeFlow:
@sbidoul Thanks for your your tips. I think now it works. Please, can you check again?

legalsylvain:
@MiquelRForgeFlow could you fix conflict ?

bealdav:
l love the idea of this PR. Any hope to be merged in September ?

MiquelRForgeFlow:
@legalsylvain done

legalsylvain (changes requested):
The ``OCABOT_USAGE`` should be updated.

https://github.com/OCA/oca-github-bot/blob/master/src/oca_github_bot/config.py#L110

bealdav:
I don't know if it's in the intention of this PR but it could be a good option if author of the PR could apply itself this rebase cmd. But not mandatory

MiquelRForgeFlow:
@bealdav yes, author of PR can rebase because they can push.

legalsylvain (approved):
Hi. 

I justed deployed the last version of this PR that is fixing authentication trouble.

Tested on a demo repository, on that PR : https://github.com/grap/github-ocabot-test/pull/15

Works like a charm. Thanks a lot for this great addition @MiquelRForgeFlow !  

image


functional test / No code review.

MiquelRForgeFlow:
@sbidoul now that it has been tested, I hope we can proceed forward :)

sbidoul:
Great. I'll now re-read since it touches a security sensitive part.

sbidoul:
Thanks for testing @legalsylvain 
Looks good, now, thanks !

Issue #57, opened 2019-07-07
merge bot: autosquash does not work


sbidoul:
It needs `--interactive`  and `GIT_SEQUENCE_EDITOR=true` in environment.

sbidoul:
This issue is obsolete since #70, since rebase is not supported anymore for now, so autosquash does not even make sense. See also https://github.com/OCA/oca-github-bot/pull/75/files#r329307563 for some explanations.

PR #75, opened 2019-09-27
autosquash non-interactive

SimoRubi:
Fixes https://github.com/OCA/oca-github-bot/issues/57

I implemented what is suggested in the issue (that is the same suggested in https://stackoverflow.com/a/12395024/11534960).
I did not manage to write an automated test because what it should check (the commit history) is completely different from what the other tests made on file merge_bot now are checking (the permissions to merge).
If you have any suggestion on how to write such a test I'll be happy to add one.

I did not even manage to test it locally because I'm not very sure about what the bot should do and where..

I only managed to verify that `GIT_SEQUENCE_EDITOR=true git rebase -i --autosquash <target_branch>` does what is expected to do!

### DIFF
--- a/src/oca_github_bot/tasks/merge_bot.py
+++ b/src/oca_github_bot/tasks/merge_bot.py
@@ -1,7 +1,6 @@
 # Copyright (c) ACSONE SA/NV 2019
 # Distributed under the MIT License (http://opensource.org/licenses/MIT).
 
-import os
 import random
 import subprocess
 from enum import Enum
@@ -174,10 +173,7 @@ def _prepare_merge_bot_branch(
     elif merge_strategy == MergeStrategy.rebase_autosquash:
         # rebase the pr branch onto the target branch
         _git_call(["git", "checkout", pr_branch])
-        _git_call(
-            ["git", "rebase", "--autosquash", "-i", target_branch],
-            env=dict(os.environ, GIT_SEQUENCE_EDITOR="true"),
-        )
+        _git_call(["git", "rebase", "--autosquash", "-i", target_branch])
     # create the merge commit
     _git_call(["git", "checkout", "-B", merge_bot_branch, target_branch])
     msg = f"Merge PR #{pr} into {target_branch}\n\nSigned-off-by {username}"
--- a/environment.sample
+++ b/environment.sample
@@ -11,7 +11,7 @@
 
 # GitHub webhook secret
 GITHUB_SECRET=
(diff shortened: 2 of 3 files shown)


src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -174,10 +173,7 @@ def _prepare_merge_bot_branch(
     elif merge_strategy == MergeStrategy.rebase_autosquash:
         # rebase the pr branch onto the target branch
         _git_call(["git", "checkout", pr_branch])
-        _git_call(
-            ["git", "rebase", "--autosquash", "-i", target_branch],
-            env=dict(os.environ, GIT_SEQUENCE_EDITOR="true"),
-        )
+        _git_call(["git", "rebase", "--autosquash", "-i", target_branch])
```
sbidoul: @SimoRubi thanks for tackling this. Actually the current version of this part of the code should be correct.

The thing is, this `rebase_autosquash` merge strategy is never triggered by the current code base.
This is because I changed the default merge strategy from rebase to merge in #70, because is a more familiar default, it preserve commit signatures, etc.

rebase with autosquash should now be asked explicitly in the merge command.

If you think a rebase merge strategy (and autosquash) are desirable features, I suggest you open an issue to specify now "ocabot merge" options to enable that.

Sorry for the confusion.

sbidoul: I've updated the blog post to mention the new default merge strategy.

sbidoul:
So I close this PR.

lreficent:
@sbidoul  Maybe this is not the best place to ask, but since the explanation is here I shoot:

> rebase with autosquash should now be asked explicitly in the merge command.

Which is the command to ask the bot for that? I was not up-to-date and though it was by default and I got surprised when it did not work [here](https://github.com/OCA/manufacture/pull/424).

sbidoul:
Yes, autosquash never worked actually, and the ocabot default merge strategy is now "merge", while the "rebase" option is gone, as there were several downside to a rebase (losing commit signatures, losing GitHub link to the PR, etc).

We could reintroduce it as an option of "ocabot merge". Like for instance "ocabot merge minor squash" to create a single commit (but with which message?), or "ocabot merge minor autosquash" to run rebase + autosquash. The former is probably the most useful, given the difficulty for contributors to create proper autosquash messages.

The best is probably to open a new issue to discuss and specify it before implementing.

PR #78, opened 2019-10-01
allexport instead of exportall

SimoRubi:
source: https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html


sbidoul:
Yes :) Thanks!

Issue #65, opened 2019-08-25
temporary_clone must not change current directory


sbidoul:
This prevents running tasks in parallel.

The `temporary_clone` must yield the temporary directory instead of changing directory.
Code must be reviewed for place that rely on the current directory to be the git clone root.
In particular all `subprocess` calls must be adapted to have a proper `cwd` argument.

sbidoul:
resolved in #79

PR #79, opened 2019-10-01
temporary_clone yields directory

SimoRubi:
Closing #65

### DIFF
--- a/src/oca_github_bot/tasks/main_branch_bot.py
+++ b/src/oca_github_bot/tasks/main_branch_bot.py
@@ -16,9 +16,9 @@ _logger = getLogger(__name__)
 @switchable("gen_addons_table")
 def _gen_addons_table(org, repo, branch, cwd):
     _logger.info("oca-gen-addons-table in %s/%s@%s", org, repo, branch)
-    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit", "--addons-dir", cwd]
+    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit"]
     subprocess.check_output(
-        gen_addons_table_cmd, universal_newlines=True, stderr=subprocess.STDOUT
+        gen_addons_table_cmd, universal_newlines=True, stderr=subprocess.STDOUT, cwd=cwd
     )
 
 
--- a/src/oca_github_bot/tasks/main_branch_bot.py
+++ b/src/oca_github_bot/tasks/main_branch_bot.py
@@ -14,16 +14,16 @@ _logger = getLogger(__name__)
 
 
 @switchable("gen_addons_table")
-def _gen_addons_table(org, repo, branch):
+def _gen_addons_table(org, repo, branch, cwd):
     _logger.info("oca-gen-addons-table in %s/%s@%s", org, repo, branch)
-    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit"]
+    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit", "--addons-dir", cwd]
     subprocess.check_output(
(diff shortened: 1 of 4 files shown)


sbidoul:
Cool, thanks! Could you do some manual testing on the main bot tasks? (ocabot merge in particular)?

src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -166,22 +167,19 @@ def _user_can_merge(gh, org, repo, username, addons_dir, target_branch):
...
         pass
     elif merge_strategy == MergeStrategy.rebase_autosquash:
         # rebase the pr branch onto the target branch
-        _git_call(["git", "checkout", pr_branch])
-        _git_call(
-            ["git", "rebase", "--autosquash", "-i", target_branch],
-            env=dict(os.environ, GIT_SEQUENCE_EDITOR="true"),
-        )
+        _git_call(["git", "checkout", pr_branch], cwd=cwd)
+        _git_call(["git", "rebase", "--autosquash", "-i", target_branch], cwd=cwd)
```
sbidoul: This part of the code never runs, but... either you need to keep the `GIT_SEQUENCE_EDITOR` environment variable, or remove -i and --autosquash

SimoRubi: Hi, thanks for the review!
I'd like to keep the `--autosquash -i` options, but I have removed the `env` keyword because I don't think it is used since `_git_call` does not have such parameter.
Is it used in some way that I don't know?

src/oca_github_bot/tasks/main_branch_bot.py
```diff
@@ -14,16 +14,16 @@
 
 
 @switchable("gen_addons_table")
-def _gen_addons_table(org, repo, branch):
+def _gen_addons_table(org, repo, branch, cwd="."):
```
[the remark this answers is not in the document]
SimoRubi: Seems legit, done

SimoRubi:
> Cool, thanks! Could you do some manual testing on the main bot tasks? (ocabot merge in particular)?

I've tested using https://github.com/SimoRubi/social/pull/22 and I found out that `oca-gen-addons-table` does not have any way to pass on the directory containing the addons.
I've created https://github.com/OCA/maintainer-tools/issues/431 to address that desired behavior

src/oca_github_bot/tasks/main_branch_bot.py
```diff
@@ -14,16 +14,16 @@
...
 
 @switchable("gen_addons_table")
-def _gen_addons_table(org, repo, branch):
+def _gen_addons_table(org, repo, branch, cwd):
     _logger.info("oca-gen-addons-table in %s/%s@%s", org, repo, branch)
-    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit"]
+    gen_addons_table_cmd = ["oca-gen-addons-table", "--commit", "--addons-dir", cwd]
     subprocess.check_output(
         gen_addons_table_cmd, universal_newlines=True, stderr=subprocess.STDOUT
     )
```
sbidoul: @SimoRubi an easy fix is to pass `cwd` to `subprocess.check_output`, no need to change `oca-gen-addons-table` for that.

SimoRubi: Thanks, I've implemented the fix.
I'll leave the issue in maintainer tools so that `oca-gen-addons-table` can be more aligned with the other methods, anyway it will be discussed in the issue itself.

sbidoul:
Ok, thanks. LGTM now. I'll merge when I have some availability ahead of me to catch and fix any possible issue quickly. Thanks for this work!

PR #91, opened 2019-11-08
[FIX] Don't duplicate maintainer mentions

dreispt:
If the the Pull Request touches several modules sharing the same maintainer, it would repeat mentions.
This change ensures that each person is mentioned only once.

Example in the wild: https://github.com/OCA/l10n-chile/pull/46#issuecomment-551707048


sbidoul:
Thanks! /ocabot merge

PR #96, opened 2019-12-26
Do not fail on twine check if an addon has no setup.py

sbidoul:
Typical example: `server_environment_file_sample`


sbidoul:
This should resolve the last recurring twine check errors in the bot.

Thanks @pedrobaeza for all the rst fixing work!

pedrobaeza:
Thanks to you for hunting down this.

PR #97, opened 2019-12-27
[ADD] ocabot migration issue maintainer

pedrobaeza:
New ocabot command for maintaining migration issues:

``/ocabot migration``, followed by the module name, performing the following:

* Look for an issue in that repository with the name "Migration to version  ``{version}``", where ``{version}`` is the name of the target branch.
* Add or edit a line in that issue, linking the module to the pull request  (PR) and the author of it.
* TODO: When the PR is merged, the line gets ticked.
* Put the milestone corresponding to the target branch in the PR.

cc @Tecnativa TT21199

### DIFF
--- a/src/oca_github_bot/tasks/migration_issue_bot.py
+++ b/src/oca_github_bot/tasks/migration_issue_bot.py
@@ -1,0 +1,113 @@
+# Copyright 2019 Tecnativa - Pedro M. Baeza
+# Copyright 2021 Tecnativa - Víctor Martínez
+# Distributed under the MIT License (http://opensource.org/licenses/MIT).
+
+import re
+
+from .. import github
+from ..config import switchable
+from ..manifest import user_can_push
+from ..process import check_call
+from ..queue import getLogger, task
+
+_logger = getLogger(__name__)
+
+
+def _create_or_find_branch_milestone(repo, branch):
+    for milestone in repo.milestones():
+        if milestone.title == branch:
+            return milestone
+    return repo.create_milestone(branch)
+
+
+def _find_issue(repo, milestone, target_branch):
+    issue_title = f"Migration to version {target_branch}"
+    issue = False
+    for i in repo.issues(milestone=milestone.number):
+        if i.title == issue_title:
+            issue = i
+            break
+    return issue
+
+
+def _set_lines_issue(gh_pr, issue, module):
+    lines = []
+    added = False
+    module_list = False
+    new_line = f"- [ ] {module} - By @{gh_pr.user.login} - #{gh_pr.number}"
+    for line in issue.body.split("\n"):
+        if line.startswith(f"- [ ] {module}"):
+            lines.append(new_line)
+            continue
+        elif not added:
+            splits = re.split(r"- \[ \] ([0-9a-zA-Z_]*)", line)
+            if len(splits) >= 2:
+                # Flag for detecting if we have passed already module list
+                module_list = True
+                line_module = splits[1]
+                if line_module > module:
+                    lines.append(new_line)
+                    added = True
+            elif module_list:
+                lines.append(new_line)
+                added = True
+        lines.append(line)
+    return lines
+
+
+@task()
+@switchable("migration_issue_bot")
+def migration_issue_start(org, repo, pr, username, module=None, dry_run=False):
+    with github.login() as gh:
+        gh_pr = gh.pull_request(org, repo, pr)
+        target_branch = gh_pr.base.ref
+        pr_branch = f"tmp-pr-{pr}"
+        try:
+            with github.temporary_clone(org, repo, target_branch) as clone_dir:
+                # Create merge bot branch from PR and rebase it on target branch
+                # This only serves for checking permissions
+                check_call(
+                    ["git", "fetch", "origin", f"pull/{pr}/head:{pr_branch}"],
+                    cwd=clone_dir,
+                )
+                check_call(["git", "checkout", pr_branch], cwd=clone_dir)
+                if not user_can_push(gh, org, repo, username, clone_dir, target_branch):
+                    github.gh_call(
+                        gh_pr.create_comment,
+                        f"Sorry @{username} you are not allowed to mark the addon to"
+                        f"be migrated.\n\n"
+                        f"To do so you must either have push permissions on "
+                        f"the repository, or be a declared maintainer of all "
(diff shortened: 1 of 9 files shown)


pedrobaeza:
@sbidoul can you take a look and if it looks good for you, deploy it on OCA server for doing real tests?

pedrobaeza:
I have put the command `migration`, but if you prefer another one, we can talk about it.

sbidoul:
@pedrobaeza thanks for working on this.

I'm good with the overall concept. A couple of comments though:
- I would split the setting of the milestone in another PR: that action is not specific for migrations, and can be done on all PRs. 
- I'm not 100% comfortable with the use of a *command* to trigger this action, perhaps it would seem more natural by looking at [MIG] in the subject, or looking for a Migration label? Actually, we could trigger this action on the presence of a Migration label, and have a separate action that sets the Migration label automatically if there is [MIG] in the subject?

src/oca_github_bot/tasks/migration_issue_bot.py
```diff
@@ -0,0 +1,103 @@
...
+                        # Flag for detecting if we have passed already module list
+                        module_list = True
+                        line_module = splits[1]
+                        if line_module > module:
+                            lines.append(new_line)
+                            added = True
+                    elif module_list:
+                        lines.append(new_line)
+                        added = True
+                lines.append(line)
```
sbidoul: This block that updates the body text could be easily extracted in a standalone function with unit tests.

src/oca_github_bot/tasks/migration_issue_bot.py
```diff
@@ -0,0 +1,103 @@
...
+                        f"To do so you must either have push permissions on "
+                        f"the repository, or be a declared maintainer of all "
+                        f"modified addons.\n\n"
+                        f"If you wish to adopt an addon and become it's "
+                        f"[maintainer]"
+                        f"(https://odoo-community.org/page/maintainer-role), "
+                        f"open a pull request to add "
+                        f"your GitHub login to the `maintainers` key of its "
+                        f"manifest.",
+                    )
```
sbidoul: Is there a persmission concern here? I mean anyone can create a migration PR and it can be added to the list automatically. Actually we could even look at the list of new modules in the PR (with a variant of `git_modified_addons` that returns new addons only), then find the migration issue and link/add modules to it.

pedrobaeza: Yes, there is. Not everyone must be able to do it, because people can hijack otherwise valid migration PRs with new ones, so PSCs/maintainers must decide, and not any other.

pedrobaeza:
> I would split the setting of the milestone in another PR: that action is not specific for migrations, and can be done on all PRs.

Well, I was thinking in being activated manually with this because not all repos follow this milestone way (for example, you with OCA/mis-builder).

> I'm not 100% comfortable with the use of a command to trigger this action, perhaps it would seem more natural by looking at [MIG] in the subject, or looking for a Migration label? Actually, we could trigger this action on the presence of a Migration label, and have a separate action that sets the Migration label automatically if there is [MIG] in the subject?

Apart from the reason exposed for checking permissions, there are incorrect migration PRs that shouldn't be pointed until a reasonable quality state is got (I did this for example until commit history is preserved at least). The last reason is that almost nobody respect commit message format, so it's almost impossible to deduce module from this title. Editing the title to conform the specs is worse as:

- It requires permission to do it (only persons with write permissions in repo can do it), so you leave out module maintainers.
- It requires to know a more difficult syntax than the command one.

pedrobaeza:
@sbidoul what do you think about my replies?

sbidoul:
@pedrobaeza I understand your argument for making it a command. Makes sens.

Splitting the setting of the milestone in another PR that is independent of this command sounds important. It can simply work by looking at the target branch and setting the milestone or label if one exist with the same name.

Also a couple of unit tests for the inner part of the command will make future maintenance much less stressful :)

sbidoul:
My last https://github.com/OCA/oca-github-bot/pull/97#issuecomment-583141924 is still current. If someone can do this split and add a small test for the inner part that should be good.

pedrobaeza:
For me the split is not needed and the test is a picky thing. This is working as is.

pedrobaeza:
I don't know how to make tests of this, and simply the setup of all the needed parts is something I can't afford, so all yours.

src/oca_github_bot/tasks/migration_issue_bot.py
sbidoul: @pedrobaeza I'm not asking for a complete integration test. I'm just asking to extract this inner part that does the updating in a function with a couple of unit tests.

pedrobaeza: OK, I will check.

pedrobaeza:
@victoralmau can you please catch this one and finish it?

victoralmau:
Hi @sbidoul i have added some test about these new command.
It's not possible to test it (green) because it's necessary to github access. What idea propose you to solve it?

legalsylvain:
@victoralmau : could you rebase ? there is a conflict.

@sbidoul : could you answer to the last question of @victoralmau ?

thanks !

sbidoul:
@victoralmau I see you started using VCR for the tests, and you have factored out the inner code for easier testing. That sounds like the right approach to me. What problem do you have ?

src/oca_github_bot/tasks/migration_issue_bot.py
ypapouin: I agree about the PR hijack, but why not letting any user add a module or link it with a PR if none is set ?

pedrobaeza: I would agree to let the maintainers do it, but not anyone. See what has happened with your confusion about the split of OCA/hr into OCA/hr, OCA/hr-holidays, OCA/hr-expense, etc. It can become very contaminated. And having this tool, it will be very easy to maintainers to raise the message to do it (or even me, that I'm doing it manually for a lot of them). Anyways, I only do it when CI is green and a few rules are respected (like preserving commit history), and that's why again I don't think this should be done by anyone.

pedrobaeza:
@sbidoul any news about this? It would be very interesting to have it for the migrations to v15 for saving a lot of maintenance work.

sbidoul:
I just deployed this. Thanks for your work!

Issue #43, opened 2019-06-17
merge bot: allow to squash commits of the PR


pedrobaeza:
Include a command for squashing the commits of the pull request independently from the fixup! message in them (not everyone makes this properly).

sbidoul:
Add an explicit command in ocabot merge command to request commit squashing.

Two options for the commit message:
1. use the PR title + description (my preference)
2. concatenate all commit messages

simahawk:
would be nice to support:
* autosquash - only for fixup! commits, preserve messages (eg: migration PR)
* squash - squash them all, use PR title as main message

legalsylvain:
Hi all. Thanks @dreispt for your answer. I'm motived to try to work on that topic. I'll keep you in touch.

guewen:
Maybe as a first step, could we have the bot refuses to merge if it finds any commit with `fixup!` and asking the contributor to squash? It could prevent having fixup commits merged such as in https://github.com/OCA/wms/pull/32

yajo:
@sbidoul a kind remainder about this issue. According to your last movements trying to improve the landing experience of functional people writing docs, this is gonna be a problem. They will probably not write correct commit message, or understand why they should (actually this happens with many devs too, after all...).

Maybe fixing this issue should land in your list of boxes to tick for that goal.

PR #103, opened 2020-03-23
[ADD] extra arguments for 3 commands oca-gen-addons-table, oca-gen-addon-readme, oca-gen-addon-icon

legalsylvain:
WIP.

Add configuration options in environment file

### DIFF
--- a/newsfragments/103.feature
+++ b/newsfragments/103.feature
@@ -1,0 +1,1 @@
+Add three new settings available in the ``environment`` file that allow to add extra argument, when calling the libraries ``oca-gen-addons-table``, ``oca-gen-addon-readme`` and ``oca-gen-addon-icon``.
--- a/src/oca_github_bot/process.py
+++ b/src/oca_github_bot/process.py
@@ -12,7 +12,9 @@ def call(cmd, cwd):
     return subprocess.call(cmd, cwd=cwd)
 
 
-def check_call(cmd, cwd, log_error=True):
+def check_call(cmd, cwd, log_error=True, extra_cmd_args=False):
+    if extra_cmd_args:
+        cmd += extra_cmd_args
     cp = subprocess.run(
         cmd,
         universal_newlines=True,
--- a/environment.sample
+++ b/environment.sample
@@ -40,3 +40,13 @@ ODOO_PASSWORD=
 # Root of the PEP 503 simple index where wheels are published
 # (publishing disabled if empty)
 SIMPLE_INDEX_ROOT=/app/run/simple-index
+
+# Space separated list of extra arguments, for the calls of the
+# following command
+# * oca-gen-addons-table
+# * oca-gen-addon-readme
+# * oca-gen-addon-icon
+# Exemple : GEN_ADDON_README_EXTRA_ARGS=--no-gen-html
+#GEN_ADDONS_TABLE_EXTRA_ARGS=
+#GEN_ADDON_README_EXTRA_ARGS=
+#GEN_ADDON_ICON_EXTRA_ARGS=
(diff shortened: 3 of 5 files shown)


src/oca_github_bot/process.py
```diff
@@ -12,7 +12,9 @@ def call(cmd, cwd):
     return subprocess.call(cmd, cwd=cwd)
 
 
-def check_call(cmd, cwd, log_error=True):
+def check_call(cmd, cwd, log_error=True, extra_cmd=False):
+    if extra_cmd:
+        cmd += extra_cmd.split(",")
```
sbidoul: I would put this split in configuration parsing. So here assume extra_cmd is a list.

sbidoul: call this `extra_cmd_args`?

sbidoul: splitting on whitespace is probably more natural (and we can consider arguments containing spaces are not supported at the moment)

legalsylvain: > I would put this split in configuration parsing. So here assume extra_cmd is a list.

Indeed.

> call this extra_cmd_args?

Indeed also.

> splitting on whitespace is probably more natural

It was to be consistent with the other args, but you're right. Space are not valid in an arg, and comma could be.

legalsylvain:
Hi @sbidoul 

1) First thanks for this great tool ! I began to test it, to have the possibility to use it for GRAP repositories, to use same merge workflow as for OCA. I'm not very comfortable with docker tools and all the technologies used (``redis``, ``flower`` but I reached to make this tools working in a reasonable time. So thanks ! #AnotherGreatToolByAcsoneTeam.

2) Regarding this PR, I was thinking to move the folllowing hardcoded values in the ``environment`` file also : ``GITHUB_STATUS_IGNORED``, ``GITHUB_CHECK_SUITES_IGNORED``. [Ref](https://github.com/OCA/oca-github-bot/blob/master/src/oca_github_bot/config.py#L57)
What do you think ?

3) I was thinking to let also the possibility to rename the bot. Ie : instead of using ``\ocabot`` have the possibility to use my [new bot](https://github.com/grap/github-ocabot-test/pull/4#issuecomment-602726178) with ``\grapbot``. More cosmetic point, but it could be great.

4) A little bit out of this topic, but I was thinking to add the possibility (in another PR), when launching ``\ocabot merge`` to have an extra argument (like ``minor``, ``patch``, etc...) named ``squach-fixup`` (not set by default). What do you think ? A lot of time, when I make PR against OCA repository, Once done, I just add fixup commits, to make the review more easy for reviewers. (not all the code to review, only the diff). but finally, when all is ok, I have to return on my PR to squash all the fixup! commits, that is a waste of time.

sbidoul:
Hi @legalsylvain , Thanks for the kind feedback. Your suggestions make sense. Can you open separate issues to track them? Regarding squashing there is #43 already.

legalsylvain:
Hi @sbidoul. It should be ok now.
it works in production, on GRAP. [test](https://github.com/grap/grap-odoo-incubator/commit/24985aa8d34aca649564f4658349d3c12ba0db74#diff-0e3720146cc8b3eaf2f9fecd4591bc23).

src/oca_github_bot/process.py
```diff
@@ -12,7 +12,9 @@ def call(cmd, cwd):
     return subprocess.call(cmd, cwd=cwd)
 
 
-def check_call(cmd, cwd, log_error=True):
+def check_call(cmd, cwd, log_error=True, extra_cmd_args=False):
```
sbidoul:
```suggestion
def check_call(cmd, cwd, log_error=True, extra_cmd_args=None):
```

False for this usage is an Odooism :)

legalsylvain: ;-)

legalsylvain:
@sbidoul : Done. (my first towncrier fragment!)

newsfragments/103.feature
```diff
@@ -0,0 +1,10 @@
+**Features**
+
```
sbidoul: The Features section be added by `towncrier`, so it must not be here. Also, towncrier will transform each newsfragment into a bullet list item, and append a link to the related PR/issue. So this should really be a simple paragraph, and if there is more to say, it should go into the documentation.

legalsylvain:
Hi @sbidoul. Thanks for your advices. Fragment updated. Better ?

sbidoul:
Perfect, thanks!

PR #107, opened 2020-03-26
[IMP] make APPROVALS_REQUIRED configurable

legalsylvain:
For the time being, APPROVALS_REQUIRED is hardcoded with the value 2. 

With that PR, it is configurable. 

Note : It is not the perfect design, as the approvals number is dynamic, in the real (OCA) life, and depends on the state of the modules. (stable / Beta / etc...) but it is a little step forward. 

Make also MIN_PR_AGE configurable.

### DIFF
--- a/environment.sample
+++ b/environment.sample
@@ -31,6 +31,8 @@ ODOO_PASSWORD=
 
 # Number of approvals to have the proposal marked as "Approved"
 #APPROVALS_REQUIRED=2
+# Number of days before the proposal can be marked as "Approved"
+#MIN_PR_AGE=5
 
 # Coma separated list of task to run
 # By default all configured tasks are run.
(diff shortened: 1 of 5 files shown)


legalsylvain:
Ok ! Thanks for the review !  A little out  of topic but I was thinking to the interest to implement two levels if the approval is green (comes from maintainers of the repo) or white (comes from any people) do you think it's relevant ? Usefull for the OCA ?

sbidoul:
> two levels if the approval is green (comes from maintainers of the repo) or white (comes from any people)

Could be useful yes. To determine who is a green approver, we could use `user_can_push` from #71. A separate PR could first refactor `_user_can_merge` into `user_can_push`.

legalsylvain:
Hi @sbidoul 

- fragment added.
- I will do another PR for adding a ``MAINTAINERS_APPROVALS_REQUIRED`` coming soon. you can proceed with this one, if you think all is OK.

PR #109, opened 2020-03-28
[REF] config and environment.sample files

legalsylvain:
This PR provides a refactoring of ``config.py`` and ``environment.sample`` files.

When I discovered this project, I was quite lost between that two files because some things was in one or in the other, (or in both). So this PR is an attempt to make a little clean and remove duplicated code or comment.

**Before this PR :**
1) some default values was present in the ``environment.sample`` file, other in the ``config.py`` file, other in both.

2) Some description was present in both files like for ``BOT_TASKS`` value, and inconsistent between two files (keys was missing) : 
Description 1 : https://github.com/OCA/oca-github-bot/blob/master/src/oca_github_bot/config.py#L50
Description 2 : https://github.com/OCA/oca-github-bot/blob/master/environment.sample#L37

3) ``GITHUB_STATUS_IGNORED`` and ``GITHUB_CHECK_SUITES_IGNORED`` was hardcoded in ``config.py`` file, and so not configurable.

4) ``BOT_TASKS`` was enumerating the tasks to run. So during an upgrade of the lib, if the value was ``all``, the new features was enabled by default (without changing the ``environment`` file) but if the value was not ``all`` the new features was disabled by default, that is not consistent. So ``BOT_TASKS`` is replace by ``BOT_TASKS_IGNORED`` assuming that most of the time, only some features are disabled and we want to have new features enabled by default.

**Important Note :**
Once merged (if accepted ;-)), the custom ``environment`` file should be updated for the existing bots that are running. (At least the OCA one) adding the following keys. (if not still defined with a custom values.)
```
HTTP_PORT=8080
APPROVALS_REQUIRED=2
MIN_PR_AGE=5
GITHUB_STATUS_IGNORED=ci/runbot,codecov/project,codecov/patch,coverage/coveralls
GITHUB_CHECK_SUITES_IGNORED=Codecov
# If current BOT_TASKS value is "all"
BOT_TASKS_IGNORED=
# Otherwise, set the list of the tasks to skip
BOT_TASKS_IGNORED=task_1,task_2
```

### DIFF
--- a/environment.sample
+++ b/environment.sample
@@ -42,6 +42,12 @@ ODOO_PASSWORD=
 #  merge_bot,merge_bot_towncrier,tag_needs_review
 #BOT_TASKS=all
 
+# Coma separated list of github status to ignore
+#GITHUB_STATUS_IGNORED=ci/runbot,codecov/project,codecov/patch,coverage/coveralls
+
+# Coma separated list of github Check suites to ignore
+#GITHUB_CHECK_SUITES_IGNORED=Codecov
+
 # Root of the PEP 503 simple index where wheels are published
 # (publishing disabled if empty)
 SIMPLE_INDEX_ROOT=/app/run/simple-index
--- a/environment.sample
+++ b/environment.sample
@@ -42,6 +42,9 @@ ODOO_PASSWORD=
 #  merge_bot,merge_bot_towncrier,tag_needs_review
 #BOT_TASKS=all
 
+# Coma separated list of task to ignore
+#BOT_TASKS_DISABLED=
+
 # Coma separated list of github status to ignore
 #GITHUB_STATUS_IGNORED=ci/runbot,codecov/project,codecov/patch,coverage/coveralls
 
--- a/environment.sample
+++ b/environment.sample
@@ -39,7 +39,7 @@ ODOO_PASSWORD=
 # Available tasks:
 #  delete_branch,tag_approved,tag_ready_to_merge,gen_addons_table,
 #  gen_addons_readme,gen_addons_icon,setuptools_odoo,mention_maintainer,
-#  merge_bot,tag_needs_review
+#  merge_bot,merge_bot_towncrier,tag_needs_review
 #BOT_TASKS=all
 
 # Root of the PEP 503 simple index where wheels are published
--- a/src/oca_github_bot/config.py
+++ b/src/oca_github_bot/config.py
@@ -16,7 +16,9 @@ def switchable(switch_name=None):
             if switch_name is None:
                 sname = func.__name__
 
-            if BOT_TASKS != ["all"] and sname not in BOT_TASKS:
+            if (
+                BOT_TASKS != ["all"] and sname not in BOT_TASKS
+            ) or sname in BOT_TASKS_DISABLED:
                 _logger.debug("Method %s skipped (Disabled by config)", sname)
                 return
             return func(*args, **kwargs)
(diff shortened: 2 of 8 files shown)


src/oca_github_bot/config.py
```diff
@@ -27,7 +27,7 @@ def func_wrapper(*args, **kwargs):
 
 
 HTTP_HOST = os.environ.get("HTTP_HOST")
-HTTP_PORT = int(os.environ.get("HTTP_PORT") or "8080")
+HTTP_PORT = int(os.environ.get("HTTP_PORT"))
```
sbidoul: Why changing this? It's good to have a default value.

src/oca_github_bot/config.py
```diff
@@ -47,12 +47,7 @@ def func_wrapper(*args, **kwargs):
 
 DRY_RUN = os.environ.get("DRY_RUN", "").lower() in ("1", "true", "yes")
 
-# Coma separated list of task to run
-# By default all configured tasks are run.
-# Available tasks:
-#  delete_branch,tag_approved,tag_ready_to_merge,gen_addons_table,
-#  gen_addons_readme,gen_addons_icon,setuptools_odoo,merge_bot,tag_needs_review
-BOT_TASKS = os.environ.get("BOT_TASKS", "all").split(",")
+BOT_TASKS_IGNORED = os.environ.get("BOT_TASKS_IGNORED", "").split(",")
```
sbidoul: I see why providing tasks to ignore can be useful. However explicitly enumerating tasks to activate is useful too, to avoid having new tasks activated by default during an upgrade. I guess we can have both, with BOT_TASK_IGNORED coming after BOT_TASKS in the config loading and deactivating some. BOT_TASK_DISABLED might be a better name BTW.

environment.sample
```diff
@@ -30,9 +30,9 @@ ODOO_PASSWORD=
 #SENTRY_DSN=
 
 # Number of approvals to have the proposal marked as "Approved"
-#APPROVALS_REQUIRED=2
+APPROVALS_REQUIRED=2
 # Number of days before the proposal can be marked as "Approved"
-#MIN_PR_AGE=5
+MIN_PR_AGE=5
```
sbidoul: having a sample file that contains default value commented out is good practice. So I'd keep that.

src/oca_github_bot/config.py
legalsylvain: OK.  Just a question, did you misspelled, or do you really want ``BOT_TASK_DISABLED`` without ``S``. For the time being, the other variable is plural ``BOT_TASKS``

legalsylvain:
Hi @sbidoul. Thanks for your review. In retrospect, I think that my first proposal was a little too radical ! 
I took out 2 of the 5 commits.
I refactored 2 commits
- regarding BOT_TASKS. (keep BOT_TASKS and add BOT_TASKS_DISABLED).
- making GITHUB_STATUS_IGNORED and GITHUB_CHECK_SUITES_IGNORED configurable

sbidoul:
LGTM, thanks! Ideally this should be in two different PRs, if only to avoid conflating two independent topics in the same newsfragment.

legalsylvain:
> LGTM, thanks! Ideally this should be in two different PRs, if only to avoid conflating two independent topics in the same newsfragment.

feature 1) ``GITHUB_STATUS_IGNORED`` and ``GITHUB_CHECK_SUITES_IGNORED`` configurable
--> available here : https://github.com/OCA/oca-github-bot/pull/111

feature 2) Add ``BOT_TASKS_IGNORED``
--> available here : https://github.com/OCA/oca-github-bot/pull/112

PR #110, opened 2020-03-29
[IMP] add nobump option for 'merge' command. Make bumpversion_mode re…

### LINKED ISSUE #90: Add nobump option to ocabot merge
We tend to see maintainers doing ocabot merge without bumping version (patch|minor|major), simply because they forget or don't know it's possible. 

In the current OCA workflow, bumping version is the most common operation.

I therefore propose to change the ocabot merge command to *require* (patch|minor|major|nobump), so the action to take in that regard is explicit and we don't have a default that does not do what we want most of the time.

legalsylvain:
- Add a fourth key ``nobump`` for the option ``bumpversion_mode`` of the command ``merge``
- Make ``bumpversion_mode`` required. (close #90)

### DIFF
--- a/src/oca_github_bot/commands.py
+++ b/src/oca_github_bot/commands.py
@@ -21,18 +21,23 @@ BOT_COMMAND_RE = re.compile(
 )
 
 
-class InvalidCommandError(Exception):
+class CommandError(Exception):
+    pass
+
+
+class InvalidCommandError(CommandError):
     def __init__(self, name):
         super().__init__(f"Invalid command: {name}")
 
 
-class OptionsError(Exception):
+class OptionsError(CommandError):
     pass
 
 
 class InvalidOptionsError(OptionsError):
     def __init__(self, name, options):
-        super().__init__(f"Invalid options for command {name}: {options}")
+        options_text = " ".join(options)
+        super().__init__(f"Invalid options for command {name}: {options_text}")
 
 
 class RequiredOptionError(OptionsError):
--- a/src/oca_github_bot/commands.py
+++ b/src/oca_github_bot/commands.py
@@ -26,9 +26,22 @@ class InvalidCommandError(Exception):
         super().__init__(f"Invalid command: {name}")
 
 
-class InvalidOptionsError(Exception):
+class OptionsError(Exception):
+    pass
+
+
+class InvalidOptionsError(OptionsError):
     def __init__(self, name, options):
         super().__init__(f"Invalid options for command {name}: {options}")
+
+
+class RequiredOptionError(OptionsError):
+    def __init__(self, name, option, values):
+        values_text = ", ".join(values)
+        super().__init__(
+            f"Required option {option} for command {name}.\n"
+            f"Possible values : {values_text}"
+        )
 
 
 class BotCommand:
@@ -53,24 +66,22 @@ class BotCommand:
 
 
 class BotCommandMerge(BotCommand):
-    bumpversion_mode = None  # optional str: major|minor|patch
+    bumpversion_mode = None
+    bumpversion_mode_list = ["major", "minor", "patch", "nobump"]
 
     def parse_options(self, options):
         if not options:
-            return
-        if len(options) == 1 and options[0] in ("major", "minor", "patch"):
+            raise RequiredOptionError(
+                self.name, "bumpversion_mode", self.bumpversion_mode_list
+            )
+        if len(options) == 1 and options[0] in self.bumpversion_mode_list:
             self.bumpversion_mode = options[0]
         else:
             raise InvalidOptionsError(self.name, options)
 
     def delay(self, org, repo, pr, username, dry_run=False):
         merge_bot.merge_bot_start.delay(
-            org,
-            repo,
-            pr,
-            username,
-            bumpversion_mode=self.bumpversion_mode,
-            dry_run=False,
+            org, repo, pr, username, self.bumpversion_mode, dry_run=False
         )
 
 
--- a/src/oca_github_bot/version_branch.py
+++ b/src/oca_github_bot/version_branch.py
@@ -34,7 +34,7 @@ def parse_merge_bot_branch(branch):
     mo = MERGE_BOT_BRANCH_RE.match(branch)
     bumpversion_mode = mo.group("bumpversion_mode")
     if bumpversion_mode == "no":
-        bumpversion_mode = None
+        bumpversion_mode = "nobump"
     return (
         mo.group("pr"),
         mo.group("target_branch"),
--- a/src/oca_github_bot/webhooks/on_command.py
+++ b/src/oca_github_bot/webhooks/on_command.py
@@ -1,8 +1,10 @@
 # Copyright (c) initOS GmbH 2019
 # Distributed under the MIT License (http://opensource.org/licenses/MIT).
 
-from ..commands import parse_commands
+from ..commands import CommandError, parse_commands
+from ..config import OCABOT_EXTRA_DOCUMENTATION, OCABOT_USAGE
 from ..router import router
+from ..tasks.add_pr_comment import add_pr_comment
 
 
 @router.register("issue_comment", action="created")
@@ -16,5 +18,18 @@ async def on_command(event, gh, *args, **kwargs):
     username = event.data["comment"]["user"]["login"]
     text = event.data["comment"]["body"]
 
-    for command in parse_commands(text):
-        command.delay(org, repo, pr, username)
+    try:
+        for command in parse_commands(text):
+            command.delay(org, repo, pr, username)
+    except CommandError as e:
+        # Add a comment on the current PR, if
+        # the command were misunderstood by the bot
+        add_pr_comment.delay(
+            org,
+            repo,
+            pr,
+            f"Hi @{username}. Your command failed:\n\n"
+            f"``{e}``.\n\n"
+            f"{OCABOT_USAGE}\n\n"
+            f"{OCABOT_EXTRA_DOCUMENTATION}",
+        )
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -6,6 +6,8 @@ import pytest
 from oca_github_bot.commands import (
     InvalidCommandError,
     InvalidOptionsError,
+    OptionsError,
+    RequiredOptionError,
     parse_commands,
 )
 
@@ -23,10 +25,10 @@ def test_parse_command_multi():
                 /ocabot merge major
                 /ocabot   merge   patch
                 /ocabot merge patch
-                /ocabot merge, please
+                /ocabot merge nobump, please
                 /ocabot merge  minor, please
                 /ocabot merge minor, please
-                /ocabot merge.
+                /ocabot merge nobump.
                 /ocabot merge patch. blah
                 /ocabot merge minor # ignored
                 ...
@@ -37,10 +39,10 @@ def test_parse_command_multi():
         ("merge", ["major"]),
         ("merge", ["patch"]),
         ("merge", ["patch"]),
-        ("merge", []),
+        ("merge", ["nobump"]),
         ("merge", ["minor"]),
         ("merge", ["minor"]),
-        ("merge", []),
+        ("merge", ["nobump"]),
         ("merge", ["patch"]),
         ("merge", ["minor"]),
     ]
@@ -50,11 +52,11 @@ def test_parse_command_2():
     cmds = list(
         parse_commands(
             "Great contribution, thanks!\r\n\r\n"
-            "/ocabot merge\r\n\r\n"
+            "/ocabot merge nobump\r\n\r\n"
             "Please forward port it to 12.0."
         )
     )
-    assert [(cmd.name, cmd.options) for cmd in cmds] == [("merge", [])]
+    assert [(cmd.name, cmd.options) for cmd in cmds] == [("merge", ["nobump"])]
 
 
 def test_parse_command_merge():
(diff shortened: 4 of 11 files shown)


tests/test_commands.py
```diff
@@ -70,12 +71,14 @@ def test_parse_command_merge():
...
+    cmds = list(parse_commands("/ocabot merge nobump"))
     assert len(cmds) == 1
     assert cmds[0].name == "merge"
-    assert cmds[0].bumpversion_mode is None
+    assert cmds[0].bumpversion_mode == "nobump"
+    with pytest.raises(RequiredOptionError):
+        list(parse_commands("/ocabot merge"))
     with pytest.raises(InvalidOptionsError):
-        list(parse_commands("/ocabot merge brol"))
+        list(parse_commands("/ocabot merge nobump BAD_OPTION"))
```
sbidoul: why changing this line?

legalsylvain: - I saw that "BAD_OPTION" was more lisible that is was wrong. But, not a big deal, I restored ``brol``.
- I added ``nobump`` because it is now required

sbidoul: > I thought we could see better that BAD_OPTION was a bad option than brol ;-)

Arguably, yes, no problem with that ;)

> ocabot merge brol

But I mean, "merge brol" was an InvalidOptionError and should still be, so let's leave that test as it is, while adding a new one for "nobump BAD_OPTION".

sbidoul:
The towncrier commit should not be in this PR.

src/oca_github_bot/commands.py
```diff
@@ -53,14 +62,20 @@ def delay(self, org, repo, pr, username, dry_run=False):
...
+    bumpversion_mode_list = ["major", "minor", "patch", "nobump"]
 
     def parse_options(self, options):
-        if not options:
-            return
-        if len(options) == 1 and options[0] in ("major", "minor", "patch"):
-            self.bumpversion_mode = options[0]
-        else:
+        # Parse bumpversion
+        res = list(set(options) & set(self.bumpversion_mode_list))
```
sbidoul: not sure I understand this line

src/oca_github_bot/commands.py
```diff
@@ -53,14 +62,20 @@ def delay(self, org, repo, pr, username, dry_run=False):
 
 
 class BotCommandMerge(BotCommand):
-    bumpversion_mode = None  # optional str: major|minor|patch
+    bumpversion_mode = None
+    bumpversion_mode_list = ["major", "minor", "patch", "nobump"]
 
     def parse_options(self, options):
-        if not options:
-            return
```
sbidoul: Why not raise RequirementOptionError here?

src/oca_github_bot/commands.py
```diff
@@ -53,14 +62,20 @@ def delay(self, org, repo, pr, username, dry_run=False):
...
-        if len(options) == 1 and options[0] in ("major", "minor", "patch"):
```
sbidoul: and adding "nobump" here?

src/oca_github_bot/commands.py
legalsylvain: because it will be raised after, if patch / minor / has not been found. no need to raise an error two times.

src/oca_github_bot/commands.py
legalsylvain: I refactored a little bit the function, with in mind that in a few times, other options will be available. (``squash`` for exemple) optional or required.

So this line is an intersection between available bumversion options and the options of the current call. It will work with ``\ocabot merge patch option_1 option_2`` and also with ``\ocabot merge option_1 option_2 patch``
I see it quite pythonistic, but if you want I can replace by a ``[x for x in options if x in self.bumpversion_mode_list]``.

src/oca_github_bot/commands.py
legalsylvain: answered above.

legalsylvain:
Hi @sbidoul thanks for your review.
In fact, the PR is still WIP. travis fails for the time being, no fragment file, etc. I just added a label, and I'll ping you when the PR is ready.

tests/test_commands.py
legalsylvain: > But I mean, "merge brol" was an InvalidOptionError and should still be, so let's leave that test as it is, while adding a new one for "nobump BAD_OPTION".

In fact no. with this refactoring, \ocabot merge bro1 is an ``InvalidOptionError`` AND a ``RequiredOptionError`` errors and it depends on the order of the check in the function.
As I refactored a little the function (see below) to let the possibility to add other arguments, now, the function check first if all the required arguments are present, and then, if there are remaining argument(s), raise an ``InvalidOptionError`` error.

legalsylvain:
@sbidoul : Now ready for review. Thanks !

src/oca_github_bot/version_branch.py
```diff
@@ -9,7 +9,7 @@
     r"-ocabot-merge"
     r"-pr-(?P<pr>\d+)"
     r"-by-(?P<username>\S+)"
-    r"-bump-(?P<bumpversion_mode>(no|patch|minor|major))"
+    r"-bump-(?P<bumpversion_mode>(nobump|patch|minor|major))"
```
sbidoul: Can we keep this unchanged? This is not necessary and would avoid issues if we upgrade the bot while merges are in progress.

Likewise I'd prefer keeping the rest of the code unchanged and modify only the command option parsing. In the code, it's good to have `bumpversion_mode=None` when we don't want to bump version, and this tranformation from "nobump" to None is easily done in the options parser.

tests/test_commands.py
```diff
@@ -70,12 +71,14 @@ def test_parse_command_merge():
...
+    cmds = list(parse_commands("/ocabot merge nobump"))
     assert len(cmds) == 1
     assert cmds[0].name == "merge"
-    assert cmds[0].bumpversion_mode is None
+    assert cmds[0].bumpversion_mode == "nobump"
+    with pytest.raises(RequiredOptionError):
+        list(parse_commands("/ocabot merge"))
     with pytest.raises(InvalidOptionsError):
-        list(parse_commands("/ocabot merge brol"))
+        list(parse_commands("/ocabot merge nobump brol"))
```
sbidoul: Ok, but you still miss a test that "ocabot merge invalidoption" (with only one invalid option) raises an error of some sort :)

legalsylvain: ``/ocabot merge invalidoption`` will not raise an invalid option, but a required option. This is tested here with ``/ocabot merge`` https://github.com/OCA/oca-github-bot/pull/110/files#diff-59262c12118a93348565a41ab7f6ce42R78

src/oca_github_bot/version_branch.py
legalsylvain: Hum. You're right. It was simplifying all the code, removing some if, but there is a risk to break some command during the process. I will roll back some changes so.
thanks for this point.

tests/test_commands.py
sbidoul: > ocabot merge invalidoption will not raise an invalid Error, but a required Error

Sorry to insist, I must be missing something: I see a test with no options, but no test with exactly one invalid option, and we had such a test before, so I think it's good to keep it.

legalsylvain: OK. Sorry, I misread your comment. I'll add it.

legalsylvain: To be complete, my approach is to test each error separately.
In the ``/ocabot merge invalidoption``, there are two errors in fact
- ``invalidoption`` is not a valid option
- ``bump_mode`` is not set. 

I was thinking that testing wich error is raised first is not interesting, and can depend on the design and change, depending on perfomance reason, or developper choice. I mean, in Odoo for exemple, you can test that create a product without name will raise an error because name is required, and test that create a product without another required field also. But know which error is raised first adds nothing (IMHO) and can be not predictive in some cases BTW.

but well : not a problem, I'll add a test ;-)
Thanks.

sbidoul: You can have both exception inherit from the same `OptionError` and test for that?

legalsylvain:
> Also don't forget newsfragments/90.feature

Side question. i was thinking to add a ``110.feature`` (the current PR) file, but you asked a ``90.feature`` (the original Issue) file. I so followed your advice ;-)
But is there some guideline in OCA documentation that says how to use ``towncrier`` ? Good practice, etc.
Not found in the [website](https://www.google.com/search?lr=&hl=fr&as_qdr=all&ei=hLWBXuWTPIq9kgWC6K_QAQ&q=towncrier+site%3Aodoo-community.org&oq=towncrier+site%3Aodoo-community.org&gs_lcp=CgZwc3ktYWIQA1DlLFi1MGCsNmgAcAB4AIABxQGIAY0FkgEDNC4ymAEAoAEBqgEHZ3dzLXdpeg&sclient=psy-ab&ved=0ahUKEwjl5ZGg68HoAhWKnqQKHQL0CxoQ4dUDCAo&uact=5) nor in [odoo-community doc repository](https://github.com/OCA/odoo-community.org/pulls?q=towncrier+).

src/oca_github_bot/commands.py
legalsylvain: ok. done.

src/oca_github_bot/commands.py
legalsylvain: also done.

legalsylvain:
Hi @sbidoul. Thanks for your help.

-> I revert the refactor of the ``parse_options()`` of the class ``BotCommandMerge``. It was an attempt to handle a more complex parsing, when there will be more options available. But I let this work for people that will implement the ``\ocabot merge minor squash`` feature. ;-)

-> I reverted also the changes that makes the branch name pattern changed. ``bump-no`` versus ``bump-nobump`` to avoid to break current jobs, when upgrading.

-> I just kept the possible values to ``["minor", "major", "patch", "nobump"]`` and I didn't wrote ``["minor", "major", "patch", None]`` list as you proposed. I see it less logic, for a required field. Hope it will not be a blocking point.

I let the fixup commit, to make more easy to read the diff. Ping me if all is OK, and if you want me to squach before merging.

src/oca_github_bot/commands.py
```diff
@@ -53,24 +66,22 @@ def delay(self, org, repo, pr, username, dry_run=False):
...
 
     def parse_options(self, options):
         if not options:
-            return
-        if len(options) == 1 and options[0] in ("major", "minor", "patch"):
+            raise RequiredOptionError(
+                self.name, "bumpversion_mode", self.bumpversion_mode_list
+            )
+        if len(options) == 1 and options[0] in self.bumpversion_mode_list:
             self.bumpversion_mode = options[0]
```
sbidoul: Nitpick I would have done this to avoid changing the rest of the code. No big deal though.

```suggestion
            if options[0] != 'nobump':
                self.bumpversion_mode = options[0]
```

legalsylvain: Yes I get it. You're right, diff could be lighter. 
The problem I see with such implementation is for developpers. IMO,

* if option is optional, the developer expect that bumpversion_mode is in ``["major", "minor", "patch", None]``
* if option is required, the developer expects that bumpversion_mode is in ``["major", "minor", "patch", "nobump"]``

I so implemented that point with such design. If it's a blocking point, I revert, no problem.

legalsylvain:
> One last question: what happen if people forget the new required "nobump" options?
If it fails silently, user will not understand why their command does nothing.
So we probably want to report a GitHub comment to avoid too much support headaches with the change.

Nice catch ! 

In fact, it's not a problem of that PR, it's a general limitation of this bot. For the time being, if we write ``ocabot make-me-a-coffee`` or ``ocabot merge`` it will fail silently. But, well, with the change of the current feature, setting an option required, it will generate a lot of mails and questions.

I so tried that changes, I set in a PR against my PR : 

- here is the code : https://github.com/legalsylvain/oca-github-bot/pull/1
- here is the result : https://github.com/grap/github-ocabot-test/pull/10

I propose you to include this feature in that current PR. It breaks "the one feature per fragment" law, but I think that it should merge together, and the second feature depends on the first one.

Do you like it ? Let me know !

sbidoul:
@legalsylvain nice error messages, it's a great improvement, thanks. I made a couple of comments on the implementation over there.

legalsylvain:
> @legalsylvain nice error messages, it's a great improvement, thanks. I made a couple of comments on the implementation over there.

Hi @sbidoul. I took into account your remarks : 

* here is the code : legalsylvain#1
* here is the result : grap/github-ocabot-test#11

kind regards.

legalsylvain:
@sbidoul. I took into account your remarks and squashed fixup! commits. could you take a look, and merge if all is OK ? Thanks !

src/oca_github_bot/webhooks/on_command.py
```diff
@@ -16,5 +19,19 @@
...
+        add_pr_comment.delay(
+            org,
+            repo,
+            pr,
+            f"Hi @{username}. Your command failed:\n\n"
+            f"``{e}``.\n\n"
+            f"{OCABOT_USAGE}\n\n"
+            f"{OCABOT_EXTRA_DOCUMENTATION}",
+        )
+        raise
```
sbidoul, line 37: No need to raise here: that will generate unnecessary noise in the logs and sentry. And these logs would not be actionable as it's a user error.

legalsylvain: done.

legalsylvain:
Hi @sbidoul. 
thanks again ! this should be OK, now.

Test : https://github.com/grap/github-ocabot-test/pull/12

legalsylvain:
Hi @sbidoul. Do you have some time to finish the review and merge this PR ?
Thanks.

sbidoul:
LGTM. Thanks for this @legalsylvain!

Now we'll need to communicate to the contributors before deploying.

legalsylvain:
Thanks ! 
I'll send an email to the OCA members.

legalsylvain:
@sbidoul : Done here : https://odoo-community.org/groups/contributors-15/contributors-160206?mode=thread

Please answer to that mail, when you update in production the ocabot.

Thanks for your time.

PR #115, opened 2020-05-20
Ignore Dependabot in check-suite

sbidoul:
The check-suites GitHub api-call started reporting Dependabot in queued state, preventing merges.
It should be safe to ignore it by default.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,7 +15,7 @@ billiard==3.5.0.4         # via celery
 bleach==3.1.4             # via readme-renderer
 celery[redis]==4.2.1
 certifi==2018.10.15       # via requests
-cffi==1.11.5              # via cryptography
+cffi==1.14.0              # via cryptography
 chardet==3.0.4            # via aiohttp, requests
 cryptography==2.3.1       # via jwcrypto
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Isn't better strategy instead of this to use a whitelist type (saying which are included) instead of a blacklist (ignoring what we don't want)?

requirements.txt
```diff
@@ -15,7 +15,7 @@ billiard==3.5.0.4         # via celery
 bleach==3.1.4             # via readme-renderer
 celery[redis]==4.2.1
 certifi==2018.10.15       # via requests
-cffi==1.11.5              # via cryptography
+cffi==1.14.0              # via cryptography
```
pedrobaeza, line 18: Shouldn't it be for compatibility with both versions?

```suggestion
cffi>=1.11.5              # via cryptography
```

sbidoul: Nope, these are the frozen dependencies for deployment.

sbidoul:
> use a whitelist

By default all checks must be green to merge. Here I'd say it's a bug with Dependabot. It should not appear as a check.

Issue #54, opened 2019-07-04
merge bot: special case bump version patch for new modules


sbidoul:
When a new module is merged, the bot currently does not bump version.
In case the commit that set the version number is not the last commit, then a .dev version goes to pypi.

To avoid this, the bot could detect that situation and force a patch version bump if the commit setting the version number is not the last one.

sbidoul:
cc/ @rousseldenis

pedrobaeza:
I see possible conflicts if the author of the PR already does the bump, isn't it?

sbidoul:
That is only for new modules (and migrations), so that should not be a problem?

pedrobaeza:
So the author puts `12.0.1.0.0`, the bot is going to put `12.0.1.0.1`?

sbidoul:
Yes, but only if the author (or the bot) made additional commits after the commit that sets 12.0.1.0.0.

pedrobaeza:
Well, you know I'm a bit frustrated by all of these weird requirements for pip deployment that are out of Odoo itself, requiring a lot of effort (no matter if it's yours or from other contributors), but let's not have again this discussion. Go ahead if it's the only solution.

sbidoul:
Well it's not only for pip. it's a way to give a better meaning to version numbers. Could be useful for the OCA apps store for instance.

pedrobaeza:
I don't see this specific case for that, as `12.0.1.0.0` is perfectly good for the rest. I suppose this is for avoiding the `--pre` problem in pypi.

sbidoul:
The real question is: which commit is 12.0.1.0.0 exactly?

pedrobaeza:
Not always the last, but that doesn't mean `12.0.1.0.0` is not valid. Several valid cases:

* The migration is done some time ago, and a reviewer ask to include a later pull request made on a previous version that applies also here. The commit is cherry-pick, but the migration is still the same.
* Migrator adds a new feature while migrating the module and decides to put it in a separate commit. `12.0.1.0.0` is still valid as there has not been previous version without that feature.

In that cases, `12.0.1.0.1` sounds ugly, as there hasn't been ever `12.0.1.0.0` version, but as said, if it's the only solution, go ahead.

emagdalenaC2i:
What happens with this type of migrations? https://github.com/OCA/server-tools/pull/1587/files Here the version is set to the same number than the fontawesome used?

sbidoul:
[...]

PR #123, opened 2020-07-02
[IMP] merge: publish wheels also in nobump mode

sbidoul:
This exception was probably done with the goal of saving space,
but for migration PRs where people use ocabot merge nobump,
we want to publish too.

### DIFF
--- a/src/oca_github_bot/tasks/merge_bot.py
+++ b/src/oca_github_bot/tasks/merge_bot.py
@@ -141,7 +141,10 @@ def _merge_bot_merge_pr(org, repo, merge_bot_branch, cwd, dry_run=False):
         d for d in modified_addon_dirs if is_addon_dir(d, installable_only=True)
     ]
 
-    # update HISTORY.rst using towncrier, before generating README.rst
+    # Update HISTORY.rst using towncrier, before generating README.rst.
+    # We don't do this if nobump is specified, because updating the changelog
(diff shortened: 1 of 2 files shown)


sbidoul:
@simahawk

src/oca_github_bot/tasks/merge_bot.py
```diff
@@ -141,7 +141,10 @@ def _merge_bot_merge_pr(org, repo, merge_bot_branch, cwd, dry_run=False):
         d for d in modified_addon_dirs if is_addon_dir(d, installable_only=True)
     ]
 
-    # update HISTORY.rst using towncrier, before generating README.rst
+    # Update HISTORY.rst using towncrier, before generating README.rst.
+    # We don't do this if nobump is specified, because updating the changelog
+    # is something we only do when "releasing", and patch|minor|major is
+    # the way to mean "release" in OCA.
```
simahawk, line 147: does this means that - say - 13.0.1.0.0 is not added to the history?

sbidoul: @simahawk Yes. There is https://github.com/OCA/oca-github-bot/issues/54 to discuss the special case of migration PRs.

simahawk:
@sbidoul can we merge this?

sbidoul:
Yep, done and deployed on the OCA server.

PR #124, opened 2020-07-25
Modified the docker file.

mimusica:
I reduced the image size by performing a better cleanup (gained ~30MB),
used the 'Linux way' of installing 3th party apps, -> the /opts folder
used the 'Elegant way' of installing python packages to their corresponding venvs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,13 +23,13 @@ RUN set -x \
   && python3 -m venv /ocamt \
   && /ocamt/bin/pip install wheel
 RUN set -x \
-  && /ocamt/bin/pip install -e git+https://github.com/OCA/maintainer-tools@73c47b6835bee3ab0eeeff7c463de6b9c085abbc#egg=oca-maintainers-tools \
+  && /ocamt/bin/pip install --no-cache-dir -e git+https://github.com/OCA/maintainer-tools@73c47b6835bee3ab0eeeff7c463de6b9c085abbc#egg=oca-maintainers-tools \
   && ln -s /ocamt/bin/oca-gen-addons-table /usr/local/bin/ \
   && ln -s /ocamt/bin/oca-gen-addon-readme /usr/local/bin/ \
   && ln -s /ocamt/bin/oca-gen-addon-icon /usr/local/bin/ \
   && ln -s /ocamt/bin/oca-towncrier /usr/local/bin/
 RUN set -x \
-  && /ocamt/bin/pip install setuptools-odoo>=2.5.0 \
+  && /ocamt/bin/pip install --no-cache-dir setuptools-odoo>=2.5.0 \
   && ln -s /ocamt/bin/setuptools-odoo-make-default /usr/local/bin/
 
 
@@ -42,7 +42,7 @@ RUN mkdir /app/tmp
 COPY ./requirements.txt /app/tmp
 RUN pip install --no-cache-dir -r /app/tmp/requirements.txt
 COPY . /app/tmp
-RUN pip install /app/tmp && rm -fr /app/tmp
+RUN pip install --no-cache-dir /app/tmp && rm -fr /app/tmp
 
 # make work and home directory
 RUN mkdir /app/run && chmod ogu+rwx /app/run
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,17 +32,13 @@ RUN set -x \
   && /ocamt/bin/pip install --no-cache-dir 'setuptools-odoo>=2.5.0' \
   && ln -s /ocamt/bin/setuptools-odoo-make-default /usr/local/bin/
 
-
 # isolate from system python libraries
 RUN python3 -m venv /app
 ENV PATH=/app/bin:$PATH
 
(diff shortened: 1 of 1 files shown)


mimusica:
The Travis tests have passed :+1:

mimusica:
@pedrobaeza Can you review? :-)

pedrobaeza:
I'm not the best suited for that. Maybe @sbidoul can do it.

yvaucher:
@mimusica you might want to check your local git config, github doesn't recognize the email and don't link it properly to your github account.

mimusica:
> @mimusica you might want to check your local git config, github doesn't recognize the email and don't link it properly to your github account.

@yvaucher Thanks for letting me know, I've modified my local user.email & user.name settings
Btw, the space gained came mostrly from '&& rm -rf /var/lib/apt/lists/*'

sbidoul:
I've several concerns:

- Doing everything in one command does not necessarily shrink the image size significantly by itself. It makes the Dockerfile harder to read, however.
- Putting the COPY statements at the top makes layer caching ineffective, so the whole change, despite reducing the image size, may actually cause more bytes to be downloaded in practice. So I'm rather -1 for that part.
- Changing /app to /opt/app, why not, but the problem with docker images is that we never know what users are doing with them, and some may depend on that directory name. So is it really important to change it?

To make the PR easier to review, it would be nice to have smaller individual changes in separate commits.
For instance, isolating the part that save most size in a separate commit would really help.

Thanks for contributing :)

sbidoul:
One other thing you could try to save space is to use `--no-cache-dir` in the other pip install commands.

mimusica:
only added '&& rm -rf /var/lib/apt/lists/*' and gained 29MB
image

mimusica:
BTW this line: '/ocamt/bin/pip install setuptools-odoo>=2.5.0'  it pipes the output of 'pip install setuptools' into a file named '2.5.0' and so it doesn't take into account the minimum version number that you want to install, which I assume isn't what you had in mind.

mimusica:
image

sbidoul:
> '/ocamt/bin/pip install setuptools-odoo>=2.5.0' it pipes the output of 'pip install setuptools' into a file named '2.5.0'

Good catch!

mimusica:
image

Dockerfile
```diff
@@ -41,7 +42,7 @@ RUN mkdir /app/tmp
 COPY ./requirements.txt /app/tmp
 RUN pip install --no-cache-dir -r /app/tmp/requirements.txt
 COPY . /app/tmp
-RUN pip install /app/tmp && rm -fr /app/tmp
+RUN pip install --no-cache-dir /app/tmp && rm -fr /app/tmp
```
sbidoul, line 45: And to shrink even more this last part could be simplified into

```Dockerfile
COPY . /app/src/oca-github-bot
RUN pip install --no-cache-dir -r /app/src/oca-github-bot/requirements.txt -e /app/src/oca-github-bot
```

so we get rid of `/app/tmp` which did hang around in layers.

mimusica:
@sbidoul what is your opinion on using    'source /ocamt/bin/activate'  & 'deactivate'  ?

sbidoul:
> using 'source /ocamt/bin/activate' & 'deactivate' ?

It should not be necessary. Using `/ocamt/bin/pip install` installs stuff in the correct virtualenv.

Dockerfile
mimusica: Do you want to remove the src (source code directory) after the installation?

yvaucher:
@sbidoul you are right, number of layers shouldn't impact the size and were important only in older version of docker as it impacted performance.

Dockerfile
sbidoul: > Do you want to remove the src (source code directory) after the installation?

No because if we install it in editable mode (-e), it is the only place where the code is available, so it is important to keep it.

Dockerfile
```diff
@@ -14,21 +14,22 @@ RUN set -x \
     python3-venv \
     rsync \
     openssh-client \
-  && apt-get clean
+  && apt-get clean \
```
yvaucher, line 17: Just from the [Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

Thus the `rm -rf /var/lib/apt/lists/*` should be enough.

Dockerfile
mimusica: I changed the file with your suggestion above and did a built, but the only size that was reduces, where the number of code lines :)

Dockerfile
mimusica: changed it

Dockerfile
mimusica: my bad, I confused it with code you have to compile (make install), but python is an interpreted language

mimusica:
Guess this is PR is finished?

mimusica:
YW

PR #127, opened 2020-09-12
[IMP] possibility to set multiple github organizations in GITHUB_ORG arg

legalsylvain:
Hi, Proposing a trivial PR to have the possibility to use a bot for multiple organizations. (I'm currently using it for two organizations).

kind regards.

### DIFF
--- a/src/oca_github_bot/config.py
+++ b/src/oca_github_bot/config.py
@@ -34,7 +34,7 @@ HTTP_PORT = int(os.environ.get("HTTP_PORT") or "8080")
 GITHUB_SECRET = os.environ.get("GITHUB_SECRET")
 GITHUB_LOGIN = os.environ.get("GITHUB_LOGIN")
 GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
-GITHUB_ORG = os.environ.get("GITHUB_ORG", "").split(",")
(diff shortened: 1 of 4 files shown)


sbidoul (changes requested):
I think this will break if GITHUB_ORG is not set.

src/oca_github_bot/config.py
```diff
@@ -34,7 +34,7 @@ def func_wrapper(*args, **kwargs):
 GITHUB_SECRET = os.environ.get("GITHUB_SECRET")
 GITHUB_LOGIN = os.environ.get("GITHUB_LOGIN")
 GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
-GITHUB_ORG = os.environ.get("GITHUB_ORG")
+GITHUB_ORG = os.environ.get("GITHUB_ORG", "").split(",")
```
sbidoul, line 37: If GITHUB_ORG is not set, this will return `['']` and then the `for org in GITHUB_ORG` wil not do what we expect.

legalsylvain: Thanks for your review.
Is it now ok ?

sbidoul: Looks good, thanks!

sbidoul (approved):

Issue #93, opened 2019-11-20
Adopt an addon campaign


sbidoul:
Following https://odoo-community.org/groups/contributors-15/contributors-150608?mode=thread&date_begin=&date_end=

See also #88

luisg123v:
What would I need to adopt a module? Just submit a PR adding myself as maintainer?

sbidoul:
> What would I need to adopt a module? Just submit a PR adding myself as maintainer?

@luisg123v yes exactly, and then make sure you perform [maintainer duties](https://odoo-community.org/page/maintainer-role).

legalsylvain:
I'm not sure to understand the workflow you propose. 
I have this one in my mind : 
- oca-github-bot scan all the repositories, and all the branches
- when he see some unmaintained modules, he opens an issue, pinging author(s) ?

sbidoul:
> workflow you propose

I imagine the bot commenting on new pull requests that touch addons without declared maintainers.

ivantodorovich:
To close, since https://github.com/OCA/oca-github-bot/pull/130 and https://github.com/OCA/oca-github-bot/pull/139 are merged?

PR #130, opened 2020-10-15
Add "adopt a module" message

eLBati:
See https://github.com/OCA/oca-github-bot/issues/93

### DIFF
--- a/environment.sample
+++ b/environment.sample
@@ -7,8 +7,8 @@
 #BROKER_URI=redis://queue
 
 # Maintainer role page URL, used in "adopt a module" message.
-# Set this empty to disable the message
-#MAINTAINER_ROLE_PAGE=https://odoo-community.org/page/maintainer-role
+# Set this to a URL that describes the maintainer role to enable the message.
+#MAINTAINER_ROLE_PAGE=
 
 #HTTP_HOST=0.0.0.0
 #HTTP_PORT=8080
--- a/environment.sample
+++ b/environment.sample
@@ -6,6 +6,10 @@
 # broker URI to use for the celery task queue
 #BROKER_URI=redis://queue
 
+# Maintainer role page URL, used in "adopt a module" message.
+# Set this empty to disable the message
+#MAINTAINER_ROLE_PAGE=https://odoo-community.org/page/maintainer-role
+
 #HTTP_HOST=0.0.0.0
 #HTTP_PORT=8080
 
--- a/src/oca_github_bot/config.py
+++ b/src/oca_github_bot/config.py
@@ -47,6 +47,10 @@ ODOO_PASSWORD = os.environ.get("ODOO_PASSWORD")
 
 BROKER_URI = os.environ.get("BROKER_URI", os.environ.get("REDIS_URI", "redis://queue"))
 
+MAINTAINER_ROLE_PAGE = os.environ.get(
+    "MAINTAINER_ROLE_PAGE", "https://odoo-community.org/page/maintainer-role"
+)
+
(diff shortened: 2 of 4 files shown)


legalsylvain (approved):
thanks ! 

could you check travis ? 

kind regards.

sbidoul:
@eLBati thanks! That was quick :+1: 

A few comments:

- A link to the maintainer role definition page would be useful in the comment. It should be configurable (https://odoo-community.org/page/maintainer-role by default).
- This feature should be behind a configuration toggle as oca-github-bot is used outside of OCA (there are many examples)
- Be careful the mention could be empty if it's a maintainer that does the PR so the implementation is probably a little bit more complex and may require some minor refactoring

eLBati:
@sbidoul I made the changes. Please review. Thanks

src/oca_github_bot/tasks/mention_maintainer.py
```diff
@@ -75,6 +78,19 @@ def get_mention(maintainers):
...
 
+def get_adopt_mention(pr_opener):
+    """Get a comment inviting to adopt the module."""
+    if config.MAINTAINER_ROLE_PAGE:
+        return (
+            "Hi @" + pr_opener + ",\n"
+            "the module you are changing has no declared maintainer.\n"
+            "Would you like to adopt it? (See %s)\n"
+            "Just add your GitHub name to `maintainers` key of `__manifest__.py`"
+            " file!"
```
sbidoul, line 89:
```suggestion
            "If you would like to adopt it, read the [maintainer role description](%s) \n"
            "and add your GitHub name to `maintainers` key of `__manifest__.py`"
            " file!"
```

environment.sample
```diff
@@ -6,6 +6,10 @@
 # broker URI to use for the celery task queue
 #BROKER_URI=redis://queue
 
+# Maintainer role page URL, used in "adopt a module" message.
+# Set this empty to disable the message
+#MAINTAINER_ROLE_PAGE=https://odoo-community.org/page/maintainer-role
```
sbidoul, line 11:
```suggestion
# Maintainer role page URL, used in "adopt a module" message.
# Set this to a URL that describes the maintainer role to enable the message.
#MAINTAINER_ROLE_PAGE=
```

src/oca_github_bot/config.py
```diff
@@ -47,6 +47,10 @@ def func_wrapper(*args, **kwargs):
 
 BROKER_URI = os.environ.get("BROKER_URI", os.environ.get("REDIS_URI", "redis://queue"))
 
+MAINTAINER_ROLE_PAGE = os.environ.get(
+    "MAINTAINER_ROLE_PAGE", "https://odoo-community.org/page/maintainer-role"
+)
```
sbidoul, line 52:
```suggestion
MAINTAINER_ROLE_PAGE = os.environ.get("MAINTAINER_ROLE_PAGE")
```

For compatibility with existing deployments.

sbidoul (commented):
Hi Lorenzo, sorry for the delay. I made a couple of suggestions to keep backward compatibility and tweak the mention a little bit. Otherwise this looks good.

sbidoul:
Oh, and I notice these maintainer mentions does not seem to be mentioned in the readme. I can add them after merging.

eLBati:
@sbidoul changes applied, thanks

sbidoul:
@eLBati I tweaked a little bit in #139 and fixed the tests. Now I think there is one last thing to do: when a maintainer adds him/herself to an addon, the adopt mention must not appear.

Issue #88, opened 2019-10-30
[FIX] do not always tag maintainer


SimoRubi:
Do not tag maintainer when:

- [x] The PR is adding the maintainer, but it's not yet added, so it shouldn't warn. https://github.com/OCA/oca-github-bot/pull/131
- [x] The maintainer is the creator of the PR, so it shouldn't warn as well.: https://github.com/OCA/oca-github-bot/pull/92
- [x] Multiple modified modules with the same maintainer.: https://github.com/OCA/oca-github-bot/pull/91

See https://github.com/OCA/project/pull/604#issuecomment-547788360

pedrobaeza:
Other case to take into account:

![Selección_005](https://user-images.githubusercontent.com/7165771/67966404-1f7f0e80-fc04-11e9-95f5-ed43623f9fe0.png)

https://github.com/OCA/brand/pull/14#issuecomment-548458594

Multiple modified modules with the same maintainer.

SimoRubi:
@OCA/tools-maintainers can someone have a look at https://github.com/OCA/oca-github-bot/pull/131?
When that one will be merged, we will finally be able to close this issue

PR #131, opened 2020-10-16
Only mention maintainers existing before the PR

SimoRubi:
Fix last part of #88: 
> The PR is adding the maintainer, but it's not yet added, so it shouldn't warn.


sbidoul:
@SimoRubi Thanks for this!

sbidoul:
Deployed at OCA

PR #151, opened 2021-08-07
Dockerfile: Install latest version of oca-gen-addons-table

sbidoul:


Dockerfile
```diff
@@ -16,20 +16,30 @@ RUN set -x \
...
-# we install them in a separate virtualenv to avoid polluting our main environment
-RUN set -x \
-  && python3 -m venv /ocamt \
-  && /ocamt/bin/pip install --no-cache-dir -U wheel
-# TODO this particular version of maintainer tool does not install with
+# The main branch bot needs several other command line tools from in OCA/maintainer-tools
+# we install them in a separate virtualenv to avoid polluting our main environment.
+
+# Install a specific version of readme and icon generator, to ensure stability
+# as any tiny change in generated output many create many commits on all addons.
```
legalsylvain, line 23: as any tiny change in generated output many create many commits on all addons.

 as any tiny change in generated output ~many~ create many commits on all addons.

PR #156, opened 2021-10-09
Support Odoo 15

sbidoul:


Dockerfile
```diff
@@ -38,11 +38,11 @@ RUN set -x \
   && python3 -m venv /ocamt \
   && /ocamt/bin/pip install --no-cache-dir -U pip wheel
 RUN set -x \
-  && /ocamt/bin/pip install --no-cache-dir -e git+https://github.com/OCA/maintainer-tools@bc8836526e40b17e56a52a28d95489b8642ef99b#egg=oca-maintainers-tools \
+  && /ocamt/bin/pip install --no-cache-dir -e git+https://github.com/OCA/maintainer-tools@d3f8cd3f3a3edded0c3eee9cbfa8274c7440516c#egg=oca-maintainers-tools \
```
pedrobaeza, line 41: Why not using master?

sbidoul: Because docker would not rebuild if that line is in the layer cache and does not change.

pedrobaeza (approved):

PR #160, opened 2021-11-13
[FIX] migration_issue_bot: Handle GH repo

pedrobaeza:
We need to pass the GH repo, not the repo name.

@Tecnativa


pedrobaeza:
Hi, @sbidoul, sorry to bother you again, but testing the command, it seems something has been lost in the way. This fix tested locally replicating the behavior now works (not sure what tests are testing though).

Can you please fast-track it and deploy it for checking if it works correctly now?

sbidoul:
@pedrobaeza done.

pedrobaeza:
Thanks! Testing and I'm afraid we need a bit more fine-tuning, but it's working in general.

PR #161, opened 2021-11-13
[FIX] migration_issue_bot: String replacements fine tuning

pedrobaeza:
- Don't replace only on the start of the string. This can lead to multiple matches on modules starting with the same name. Although the rest of the added code now avoid to make more than one replacement, it's good to still do the exact match. This is provided through a regular expression matching the module name + space or the end of the line.
- Provide line filling shortcut for faster completion.
- Preserve possible existing mark check.

@Tecnativa


pedrobaeza:
@sbidoul with this, everything should be correct and ready for a massive usage. Thanks for all.

sbidoul:
deployed

PR #162, opened 2021-11-14
[FIX] migration_issue_bot: Line ending + already checked marks

pedrobaeza:
This commit includes 2 fixes:

- Add to the regular expression that detects the line the carrier return (\r), as `$` is not representing the same as it's fed from the issue text.
- For detecting if the module list has ended, we were not taking into account that there can be modules already migrated, and thus, the `x` should be included in such regular expression.

@Tecnativa


pedrobaeza:
Hi, @sbidoul, it's me again, sorry. I've continued testing the command and found what I expect the last 2 bugs.

sbidoul:
Please, please, increase test coverage on that function. It should be easy.

sbidoul:
deployed

PR #163, opened 2021-12-06
Hide secrets when commenting on PRs

sbidoul:
Writing github bots is hard.

We happen to be leaking a secret. Example here: https://github.com/OCA/maintenance/pull/157#issuecomment-981837484
This explains why our GitHub token was revoked.


sbidoul:
deployed

Issue #122, opened 2020-06-25
Allow addon maintainers to ocabot merge migration PRs


sbidoul:
First thing is to write down the algorithm...

legalsylvain:
Hi @sbidoul : I'm very interested by this one. I don't know how to implement it however.

Use case : 
- I  developed a module named ``account_template_active`` in https://github.com/OCA/account-financial-tools (branch 12.0). [link](https://github.com/OCA/account-financial-tools/tree/12.0/account_template_active)
- Someone port this module in V14.0. (https://github.com/OCA/account-financial-tools/pull/1187)
- I'd like to merge the PR once the approval process is done.

Different approaches at the first sight : 
1. git fetch various previous branches (13.0, 12.0, 11.0, ...), until founding a branch that contain an installable module name ``account_template_active``. Check if the user that call ``ocabot merge`` is the maintainer of the latest version of the module.
2. Call github and wget ``https://github.com/OCA/account-financial-tools/blob/XX.0/account_template_active/__manifest__.py`` files and check if the manifest contains the author of the call.
3. Call via API, odoo-community.org. (not sure, it seems that somebody said me it was not possible). (BTW, it seems that the instance doesn't fetch latest modules because the module ``account_template_active`` is not available on OCA website)
4. Other ?

pedrobaeza:
As I don't know the internals of how this is done, I can't say which one is the best approach, but it seems one the third option is not very feasible. I see the number 1 as the way to go.

PR #183, opened 2022-04-13
[IMP] Search addons maintainers in other branches. Fix : #122

legalsylvain:
Search for maintainers in all the branches, to make possible maintainer to merge migration PRs. (#122)
Note, it works also for the past. (If i'm declared as maintainers of a 12.0 module, I will also have the possibility to merge PR against 8.0 (for backport / fixes / etc...)

**Deployment Note**
Once deployed, you should update your ``environment`` file. if not, the feature will be simply anavailable.

### DIFF
--- a/src/oca_github_bot/manifest.py
+++ b/src/oca_github_bot/manifest.py
@@ -267,11 +267,10 @@ def is_maintainer_other_branches(org, repo, username, modified_addons, other_bra
         is_maintainer = False
         for branch in other_branches:
             manifest_file = (
-                float(branch) < 10.0 and "__openerp__.py" or "__manifest__.py"
+                "__openerp__.py" if float(branch) < 10.0 else "__manifest__.py"
             )
             url = (
-                f"https://raw.githubusercontent.com/{org}/{repo}/{branch}/"
-                f"{addon}/{manifest_file}"
+                f"https://github.com/{org}/{repo}/raw/{branch}/{addon}/{manifest_file}"
             )
             _logger.debug("Looking for maintainers in %s" % url)
             r = requests.get(
--- a/src/oca_github_bot/manifest.py
+++ b/src/oca_github_bot/manifest.py
@@ -77,8 +77,9 @@ def get_manifest_path(addon_dir):
     return None
 
 
-def parse_manifest(manifest: bytes)-> dict:
+def parse_manifest(manifest: bytes) -> dict:
     return ast.literal_eval(manifest.decode("utf-8"))
+
 
 def get_manifest(addon_dir):
     manifest_path = get_manifest_path(addon_dir)
--- a/src/oca_github_bot/manifest.py
+++ b/src/oca_github_bot/manifest.py
@@ -77,12 +77,15 @@ def get_manifest_path(addon_dir):
     return None
 
 
+def parse_manifest(manifest: bytes)-> dict:
+    return ast.literal_eval(manifest.decode("utf-8"))
+
 def get_manifest(addon_dir):
     manifest_path = get_manifest_path(addon_dir)
     if not manifest_path:
         raise NoManifestFound(f"no manifest found in {addon_dir}")
-    with open(manifest_path, "r") as f:
-        return ast.literal_eval(f.read())
+    with open(manifest_path, "rb") as f:
+        return parse_manifest(f.read())
 
 
 def set_manifest_version(addon_dir, version):
@@ -277,7 +280,7 @@ def is_maintainer_other_branches(org, repo, username, modified_addons, other_bra
                 url, allow_redirects=True, headers={"Cache-Control": "no-cache"}
             )
             if r.ok:
-                manifest = ast.literal_eval(r.content.decode("utf-8"))
+                manifest = parse_manifest(r.content)
                 if username in manifest.get("maintainers", []):
                     is_maintainer = True
                     break
--- a/src/oca_github_bot/manifest.py
+++ b/src/oca_github_bot/manifest.py
@@ -2,9 +2,13 @@
 # Distributed under the MIT License (http://opensource.org/licenses/MIT).
 
 import ast
+import logging
 import os
 import re
 
+import requests
+
+from . import config
 from .github import git_get_current_branch, github_user_can_push
 from .process import check_call, check_output
 
@@ -16,6 +20,8 @@ BRANCH_RE = re.compile(r"^(?P<series>\d+\.\d+)$")
 MANIFEST_VERSION_RE = re.compile(
     r"(?P<pre>[\"']version[\"']\s*:\s*[\"'])(?P<version>[\d\.]+)(?P<post>[\"'])"
 )
+
+_logger = logging.getLogger(__name__)
 
 
 class NoManifestFound(Exception):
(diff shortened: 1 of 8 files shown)


sbidoul:
I think we should do this only for migration PRs, i.e. when the addon is not yet on the target branch. Otherwise, this may have security implications, because a maintainers on a branch may not be the same as on other branches.

legalsylvain:
> I think we should do this only for migration PRs, i.e. when the addon is not yet on the target branch. Otherwise, this may have security implications, because a maintainers on a branch may not be the same as on other branches.

Good question I asked myself last night !

Well, in fact, when I adopt an addon, I adopt it _most of the time_ "globally". = "I'd like to have the possibility to merge bugfixes, and backport of bugfixes in older version". 

Making a PR to add the key in the manifest file is a workflow that can be slow in some repo without a lot of reviewers, but force me to make a PR against all the branches the module is available will make the workflow slower.

typical use case :  I adopt a module in V12, (making a PR). The PR is merged, but in the meantime, the port of this module has been done for V14. I'm so not the maintainer of the V14 module and I have to make another PR to be set as maintainer. but maybe a V15 PR is pending ! and I'll have to make a third PR. Finally, when i'll make the port of the module in V16 (that is the next release I target) I'll not have the possibility to merge the PR.

> because a maintainers on a branch may not be the same as on other branches.

Is it a real use case ? I mean : "module in V12 maintained by @sbidoul  + module in V15 maintained by @legalsylvain" 
1) I'm not sure that this case exists.
2) But even if it were the case, do we really want to make sure that you are not allowed to merge a patch in V16 ?

So, After these nocturnal reflections, my point of view was: 
1) Be optimistic ! if we can make bugfix / migration workflow quicker, it's great ! And if abuse is observed, we can always change the bot's behavior very quickly. at one time when I was on the board, I had the rights to all the repos in the OCA, but still, I didn't throw any ``ocabot merge'' into repos/modules that I didn't understand.
2) I think that fundamentally, the problem lies elsewhere and should be managed more globally. Over time, people declared as maintainers of OCA modules will leave. And so, their account will be able to modify OCA code forever, which can be a more global problem.

legalsylvain:
Hello @sbidoul ! Do you keep your point of view ? I'd like to move forward with that PR. I can change the algorithm if it's a blocking point. Let me know !

sbidoul:
Sorry I had lost track of this. I think you are right. If and when we have this kind of conflicts between maintainers in different branches, we can solve that by other means. So +1 with your approach.

src/oca_github_bot/manifest.py
```diff
@@ -234,6 +246,43 @@ def user_can_push(gh, org, repo, username, addons_dir, target_branch):
...
+        org, repo, username, modified_addons, other_branches
+    )
+
+
+def is_maintainer_other_branches(org, repo, username, modified_addons, other_branches):
+    for addon in modified_addons:
+        is_maintainer = False
+        for branch in other_branches:
+            manifest_file = (
+                float(branch) < 10.0 and "__openerp__.py" or "__manifest__.py"
```
sbidoul, line 270:
```suggestion
                "__openerp__.py" if float(branch) < 10.0 else "__manifest__.py"
```

src/oca_github_bot/manifest.py
```diff
@@ -234,6 +246,43 @@ def user_can_push(gh, org, repo, username, addons_dir, target_branch):
...
+            )
+            url = (
+                f"https://raw.githubusercontent.com/{org}/{repo}/{branch}/"
+                f"{addon}/{manifest_file}"
```
sbidoul, line 274:
```suggestion
                f"https://github.com/{org}/{repo}/raw/{branch}/{addon}/{manifest_file}"
```

src/oca_github_bot/manifest.py
```diff
@@ -234,6 +246,43 @@ def user_can_push(gh, org, repo, username, addons_dir, target_branch):
...
+            )
+            _logger.debug("Looking for maintainers in %s" % url)
+            r = requests.get(
+                url, allow_redirects=True, headers={"Cache-Control": "no-cache"}
+            )
+            if r.ok:
+                manifest = ast.literal_eval(r.content.decode("utf-8"))
```
sbidoul, line 281: nit: it might be better to add a `parse_manifest(manifest: bytes)` function in `manifest.py`.

sbidoul (approved):
LGTM, a couple of suggestions.

legalsylvain (commented):
hi @sbidoul. Thanks for your review. I accepted your 2 proposal.
another comment inline.

otherwise, LGTM.

sbidoul:
:+1: If it's ready for you I can deploy to the OCA server tomorrow.

legalsylvain:
> +1 If it's ready for you I can deploy to the OCA server tomorrow.

Yes ! ready from my part.

when deploying don't forget to add 
`` MAINTAINER_CHECK_ODOO_RELEASES=8.0,9.0,10.0,11.0,12.0,13.0,14.0,15.0``
in the ``environment`` file otherwise, the feature will be disabled.

Please ping me if troubles.

legalsylvain:
> +1 If it's ready for you I can deploy to the OCA server tomorrow.

Hi. did you deployed the feature ? 

Just to know if https://github.com/OCA/product-attribute/pull/1083#issuecomment-1137313667 is normal.

sbidoul:
No, it was a sunny week-end :) I'll do it in the next days.

legalsylvain:
> No, it was a sunny week-end :) I'll do it in the next days.

hum... Next week-end will be sunny also ! ;-)

Let me know when it's done. 

regards.

sbidoul:
Hi @legalsylvain. It is deployed. I'll let you test and announce :)

legalsylvain:
Hi @sbidoul. Seems to work :   

https://github.com/OCA/product-attribute/pull/1083#issuecomment-1140407144

Thanks !

sbidoul:
Excellent. Thanks for the contrib!

sbidoul:
Can you also try merging something you are not allowed to ?

eLBati:
@legalsylvain Hi, following https://github.com/OCA/oca-github-bot/commit/108c9aea1a1a3a1055f1513ddfdb37e2fe1d7d0f#r83501261 , we would need to grant some users the ability to merge PRs only for a specific (old) version, as they have competence in that version and PSC members do not work anymore on that (old) version

legalsylvain:
Hi @elbati. Thanks for your message.

However : I understand what you ask, but I dont understand the problem you try to fix.
Do you face some abuses, people that merge PR that should not be merged ? 

Thanks for your precisions.

TheMule71:
> I dont understand the problem you try to fix.

Hi there!
Thank you for your answer. Please allow me to clarify a bit more.

There's no problem to fix. It's just a feature request.

Someone volunteered to maintain a very old branch (8.0), but they have clearly stated they are not interested in moving on from that, and they're not going to contribute anything to more recent branches.  We would like to grant them maintainer status for all modules in the 8.0 branch only.

Currently, that would imply granting maintainer status for all modules in other branches too, something completely unnecessary (per their request). We feel that limiting permissions to the minimun required is best practice anyway.

So we're wondering if there's a way to override the environtment variable `MAINTAINER_CHECK_ODOO_RELEASES=8.0,9.0,10.0,11.0,12.0,13.0,14.0,15.0`  and set it locally (for our repo) to something along the line of `MAINTAINER_CHECK_ODOO_RELEASES=10.0,11.0,12.0,13.0,14.0,15.0`.

Regarding https://github.com/OCA/oca-github-bot/pull/183#issuecomment-1099157614 we don't need module granularity at all. Excluding maintainers specified in the 8.0 branch as a whole would be fine for us. We don't need extra complexity, just a way to override the `MAINTAINER_CHECK_ODOO_RELEASES` environment variable. (Apologies if there's a documented way of doing that, if so it went completely over our heads).

PR #190, opened 2022-07-01
[IMP] make _set_lines_issue working on empty issue

legalsylvain:
Hi.

when calling ``/ocabot migration``, if the issue doesn't contains lines, the command fail silently. (that is due to the github api I guess. If we "edit" an github issue, and write the same text as before, the write is ignored).

with that patch, the line is added at the end of the migration issue body.

**Context** 
- can be used on custom modules, if we don't want to initialize the repository.
- usefull also in the Openupgrade project (See : https://github.com/OCA/OpenUpgrade/issues/3288#issuecomment-1157036504 
with the decision to not populate migration issue).

CC : @pedrobaeza, @sbidoul 

thanks for your review.

### DIFF
--- a/src/oca_github_bot/tasks/migration_issue_bot.py
+++ b/src/oca_github_bot/tasks/migration_issue_bot.py
@@ -29,12 +29,12 @@ def _find_issue(gh_repo, milestone, target_branch):
     return issue
 
 
-def _set_lines_issue(gh_pr, issue, module):
+def _set_lines_issue(gh_pr_user_login, gh_pr_number, issue_body, module):
     lines = []
     added = False
     module_list = False
-    new_line = f"- [ ] {module} - By @{gh_pr.user.login} - #{gh_pr.number}"
-    for line in issue.body.split("\n"):
+    new_line = f"- [ ] {module} - By @{gh_pr_user_login} - #{gh_pr_number}"
+    for line in issue_body.split("\n"):
         if added:  # Bypass the checks for faster completion
             lines.append(line)
             continue
@@ -62,7 +62,7 @@ def _set_lines_issue(gh_pr, issue, module):
     # make the addition working on an empty migration issue
     if not added:
         lines.append(new_line)
-    return lines
+    return "\n".join(lines)
 
 
 @task()
@@ -113,8 +113,10 @@ def migration_issue_start(org, repo, pr, username, module=None, dry_run=False):
                 )
                 return
             # Change issue to add the PR in the module list
-            lines = _set_lines_issue(gh_pr, issue, module)
-            issue.edit(body="\n".join(lines))
+            new_body = _set_lines_issue(
+                gh_pr.user.login, gh_pr.number, issue.body, module
+            )
+            issue.edit(body=new_body)
         except Exception as e:
             github.gh_call(
                 gh_pr.create_comment,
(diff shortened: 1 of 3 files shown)


HaraldPanten:
Hi @legalsylvain thank you for this work.

If I understood well, if we call the ocabot migration command in a migration PR that has not been added in the issue before, a new line will be added with the name of the module, right?

My suggestion is... Why not adding it by alphabetical order? Is it possible?

THX!

legalsylvain:
Hi @HaraldPanten

hi. thanks for your review. 
i'm not sure to understand correctly your remark.
- the current PR is about the possibility to add a line if no lines are present in the migration issue. So, as it is the first element, there is nothing to order alphabetically.
- regarding the insertion of a new line, if lines are present, it is working very well. see the historic of this PR : https://github.com/OCA/OpenUpgrade/issues/3287

src/oca_github_bot/tasks/migration_issue_bot.py
```diff
@@ -58,6 +58,10 @@ def _set_lines_issue(gh_pr, issue, module):
                 lines.append(new_line)
                 added = True
         lines.append(line)
+
+    # make the addition working on an empty migration issue
+    if not added:
+        lines.append(new_line)
```
sbidoul, line 64: A typical migration issue ends with comments after the check list of modules:

image

So this would add a line after the comment which may not look very nice.

I wish we had better tests for `_set_lines_issue`.

legalsylvain: > A typical migration issue ends with comments after the check list of modules:

that's not the point of the PR. 

The line will be added only if the issue doesn't contain module list.
In your screenshot, my patch has no effect, because previous code will add the line in the correct place, and so added will be true.

That is just a trivial patch to initialize migration issue, if there is no module list.

legalsylvain:
@sbidoul. I refactored the function to make explicit tests. (coverage: +0,5%)

Given a PR '11' made by 'sbidoul' user regarding 'mis_builder' module.

1)
```
Issue with list but not the module
- [ ] a_module_1 - By @legalsylvain - #1
- [ ] z_module_1 - By @pedrobaeza - #2
```
become 
```
Issue with list but not the module
- [ ] a_module_1 - By @legalsylvain - #1
- [ ] mis_builder - By @sbidoul - #11
- [ ] z_module_1 - By @pedrobaeza - #2
```
2)
```
Issue with list containing the module
- [x] mis_builder - By @legalsylvain - #1
- [ ] z_module_1 - By @pedrobaeza - #2
```
become
```
Issue with list containing the module
- [x] mis_builder - By @sbidoul - #11
- [ ] z_module_1 - By @pedrobaeza - #2
```

3)
```
Issue with no list
```
become 
```
Issue with no list
- [ ] mis_builder - By @sbidoul - #11
```


Note : Use case 1 & 2 are the current behaviour.
The patch of this PR (first commit) allow the Use case 3.
(without that patch, ```Issue with no list``` become ```Issue with no list```.

Let me know if it's not clear.

sbidoul (approved):
Thanks a lot for the refactoring and test, Sylvain. It's much clearer now :)
Can you also add the news fragment ?

legalsylvain:
Hi @sbidoul. I added a ``190.bugfix`` fragment.

Question : there is pending ``183.bugfix`` file that has not disappeared even if the PR #183 is merged. Is it normal ?

sbidoul:
> Question : there is pending 183.bugfix file that has not disappeared even if the PR https://github.com/OCA/oca-github-bot/pull/183 is merged. Is it normal ?

Yes, they disappear when we run towncrier at release time.

legalsylvain:
> Yes, they disappear when we run towncrier at release time.

Thanks ! I thought it was during merge process.

PR #191, opened 2022-07-01
[ADD] add warning in ocabot migration if previous PR was present for the same module

legalsylvain:
**Use case**
- When maintainers are calling ``ocabot migration`` 
- if previous PR was present, 

**Current behaviour** 
The line of the migration issue is replaced silently by the new PR. (and new author).
However, in many cases, it is a duplication of work.

**Current workaround**

Maintainers should take a look on the migration issue, to check if an existing PR still exist and alert on the PR. See for exemple : https://github.com/OCA/OpenUpgrade/pull/3334#issuecomment-1164117298
For repos with a large amount of modules, it is very complicated to remember that a porting is already in progress.

**With that PR**

If an existing PR is referenced in the migration issue, a message is added to the new PR by the bot, referencing the first PR and pinging the original author.

image

CC : 
- @sbidoul 
- @pedrobaeza, you could be interested by this one ?

Note : tested with ngrox on a grap repo. See : https://github.com/grap-org-test-bot/github-ocabot-test/pull/29#issuecomment-1172407880

### DIFF
--- a/newsfragments/191.feature
+++ b/newsfragments/191.feature
@@ -1,0 +1,1 @@
+When calling ``/ocabot migration`` if a previous Pull Request was referenced in the migration issue, post an alert on the new Pull Request to mention the previous work.
--- a/src/oca_github_bot/tasks/migration_issue_bot.py
+++ b/src/oca_github_bot/tasks/migration_issue_bot.py
@@ -33,6 +33,7 @@ def _set_lines_issue(gh_pr_user_login, gh_pr_number, issue_body, module):
     lines = []
     added = False
     module_list = False
+    old_pr_number = False
     new_line = f"- [ ] {module} - By @{gh_pr_user_login} - #{gh_pr_number}"
     for line in issue_body.split("\n"):
         if added:  # Bypass the checks for faster completion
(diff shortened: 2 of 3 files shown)


legalsylvain:
Rebase done. This one is ready to review. 

CC : @OCA/openupgrade-maintainers : this one could interest you in case two contributors work on the same module migration.

pedrobaeza:
Isn't better to not overwrite it in such case unless the PR is closed?

legalsylvain:
> Isn't better to not overwrite it in such case unless the PR is closed?

I don't know ! I abstain on that topic. Feel free to open an issue / PR if you think it's relevant. In my opinion, maintainers will have to do manual work. (if we keep old PR, maybe we should set the new one manually, and if we get the new PR, maybe we should set the old one manually...) We can not predict what is the best PR.

My concern is to make sure that everyone knows that there is a duplicate work in progress. (The author of the new PR, the author of the previous PR, and the rest maintainers). For the time being, with 130 openened PR on Openupgrade project, I really can't remember if anyone has already worked on the migration from account to V15.

Note : that PR is not changing the algorithm you introduced in #97. That is the current behaviour.

legalsylvain:
Note : CI is failing for weird reason. Don't understand.

pedrobaeza:
I know that my initial intention was to always overwrite, thinking that maintainers will know what to do, but now seeing in perspective, it seems that people don't look too much, so that's why I say to change the initial intention to prevent such change unless the PR is closed.

legalsylvain:
You probably right. I created thé issue. https://github.com/OCA/oca-github-bot/issues/194

In the meantime, could you review this one ?

Thanks !

pedrobaeza (approved):

legalsylvain:
Hi @hbrunn. 

Thanks for your review. i accepted your two changes and squashed commits.

Should be OK.

Regards.

hbrunn (approved):

sbidoul:
@legalsylvain I just deployed the master branch on the OCA server

legalsylvain:
Thanks !

PR #205, opened 2022-09-18
Fix setting erppeek context

sbidoul:


sbidoul:
The erppeek client context is documented as being None by default.

StefanRijnhart:
Thanks! I'm trying to patch up the linting here: https://github.com/OCA/oca-github-bot/pull/206

StefanRijnhart (approved):
Looks like clabot is posting messages again since you deployed the change! For example https://github.com/OCA/sale-workflow/pull/2114#issuecomment-1250608604

sbidoul:
> Looks like clabot is posting messages again since you deployed the change

Good! But of course it's again a false alarm on a migration PR :/

sbidoul:
Thanks for the linters!

PR #212, opened 2022-09-27
Refactor wheel builder

sbidoul:
- reuse the build environment for performance (and stability if/when we pin build tools versions)
- support (and give priority to) `pyproject.toml` in addon directories, towards removing the `setup` directories

TODO
- [x] test with a pyproject-enabled addon

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+asyncio_mode = strict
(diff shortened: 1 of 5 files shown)


sbidoul:
I'll merge this, deploy and monitor. This should be fully backward compatible.

PR #215, opened 2022-10-10
[FIX] ocabot migration duplicating lines in migration issue?

### LINKED ISSUE #214: ocabot migration duplicating lines in migration issue?
See for example https://github.com/OCA/server-env/issues/120 where the issue was created correctly but now lines with references to a PR are duplicated.

SirTakobi:
Fix #214

### DIFF
--- a/src/oca_github_bot/tasks/migration_issue_bot.py
+++ b/src/oca_github_bot/tasks/migration_issue_bot.py
@@ -39,7 +39,7 @@ def _set_lines_issue(gh_pr_user_login, gh_pr_number, issue_body, module):
         if added:  # Bypass the checks for faster completion
             lines.append(line)
             continue
-        groups = re.match(rf"^- \[( |x)\] {module}( |\r)", line)
+        groups = re.match(rf"^- \[( |x)\] \b{module}\b", line)
         if groups:  # Line found
             # Get the Old PR value
             regex = r"\#(\d*)"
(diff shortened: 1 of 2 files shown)


src/oca_github_bot/tasks/migration_issue_bot.py
```diff
@@ -39,7 +39,7 @@ def _set_lines_issue(gh_pr_user_login, gh_pr_number, issue_body, module):
         if added:  # Bypass the checks for faster completion
             lines.append(line)
             continue
-        groups = re.match(rf"^- \[( |x)\] {module}( |\r)", line)
+        groups = re.match(rf"^- \[( |x)\] {module}", line)
```
pedrobaeza, line 42: This is not correct at all, as it can match with partial module names. Why is it incorrect? Maybe we can match the end of the line instead.

SirTakobi: Thanks for the review!
I haven't managed to reproduce exactly the issue #214 because I don't know how the Migration issue is created, but I think this regex might be the cause.

I have found that `\b` is made exactly for matching full words, see https://docs.python.org/3/library/re.html#regular-expression-syntax, can you check if it looks better now?

pedrobaeza (approved):
Thank you for tackling the problem!

sbidoul:
Thank you @SirTakobi. I just deployed.

pedrobaeza:
Let's see if we finally end with this problem. Thanks!

Issue #159, opened 2021-11-02
[RFC] bot message when migration PR doesn't respect history.


legalsylvain:
**Event** : On new PR

**Check** : If PR name contains ``[MIG]`` (not sensitive check) AND commit quantity = 1

**Message** : 

_Thanks for the contribution._

_Please preserve commit history following technical method explained in https://github.com/OCA/maintainer-tools/wiki/#migration_

_If the jump is between several versions, you have to modify the source branch in the main command to accommodate it to this circumstance._

_You can also take a look on the project https://github.com/OCA/odoo-module-migrator/ to make easier migration_

Ref : @pedrobaeza in https://github.com/OCA/web/pull/2067#issuecomment-957251756.

Nice to have feature in my opinion to avoid to repeat always the same thing. (even it will not cover all the case, specially if the PR contains more than 1 commit).

pedrobaeza:
👍

rousseldenis:
Mmmh, your check can be false in some repositories as they maintain branches from preceding ones (and so, len(migration commit) == 1.

I'm in favor of this, but with a check a little bit cleverer (but don't know how yet).

Apart of that, I think we should change OCA main branches behaviour with the one I described above. It will facilitate review in order to have less commit history

legalsylvain:
What about to change the texte of the message ? Something like "are you sûre you preserved history" ?

legalsylvain:
@rousseldenis : any pov regarding my proposal ?

rousseldenis:
> @rousseldenis : any pov regarding my proposal ?

Yes that's sound good.

IMHO, we should add to conditions 'if module directory isn't in repo yet'. So, this will work in all conditions I explained before.

PR #220, opened 2022-10-11
[ADD] migration_pr_check: Remind migration guidelines

SirTakobi:
Possible implementation of #159: this always adds a comment about migration guidelines when the PR is identified as a _migration PR_.

The result is in https://github.com/SirTakobi/l10n-italy/pull/28

### DIFF
--- a/environment.sample
+++ b/environment.sample
@@ -84,3 +84,6 @@ OCABOT_TWINE_REPOSITORIES="[('https://pypi.org/simple','https://upload.pypi.org/
 # List of branches the bot will check to verify if user is the maintainer
 # of module(s)
 MAINTAINER_CHECK_ODOO_RELEASES=8.0,9.0,10.0,11.0,12.0,13.0,14.0,15.0
+
+# Reminder of migration guidelines.
+#MIGRATION_GUIDELINES_REMINDER=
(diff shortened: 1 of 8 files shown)


legalsylvain:
Hi @SirTakobi
just reviewed your code. LGTM. Just one question : to know if a PR is a migration one, you parse the related issue. That's great but in my opinion not complete. i think we could also guess it's a migration PR if the title contains "[MIG]"  or " mig ". (many times, the issue "Migration in version 16.0" is not updated.

What do you think ?

SirTakobi:
> Hi @SirTakobi just reviewed your code. LGTM. Just one question : to know if a PR is a migration one, you parse the related issue. That's great but in my opinion not complete. i think we could also guess it's a migration PR if the title contains "[MIG]" or " mig ". (many times, the issue "Migration in version 16.0" is not updated.
> 
> What do you think ?

Thanks for having a look!
I added a check on the PR Title with its test; it is mocking more than I'd like to, but I think there is no other way to make it work.
I also fixed a false positive in another test.

pre-commit is failing while building `isort`'s virtual Environment, I suppose it will be fixed after https://github.com/OCA/oca-github-bot/pull/229

SirPyTech:
Moved to https://github.com/OCA/oca-github-bot/pull/363.

PR #243, opened 2023-06-17
Update pinned dependencies, bump required python to 3.10

sbidoul:


sbidoul:
The pinned dependencies would be slightly different between python 3.8 and 3.9.
Also I notice the `build` tool  fails on 3.8 but works on 3.10.

So time as come to bump the minimum required python version. 

Therefore I also update the docker image to use Ubuntu 22.

cc/ @legalsylvain

legalsylvain (approved):
Sorry, I missed this one. LGTM. Thanks !

sbidoul:
Superseded by #293

PR #245, opened 2023-06-17
Fix issue with error reporting

### LINKED ISSUE #244: Some errors in the merge process are not reported
This is due to the rendering of the command in the error message, such as here

https://github.com/OCA/oca-github-bot/blob/48b34f3b2d21fa004811afab0e47d0a0c0131bcb/src/oca_github_bot/tasks/merge_bot.py#L408

when one of the command components is a `pathlib.Path`, this fails.

sbidoul:
fixes #244

### DIFF
--- a/newsfragments/244.bugfix
+++ b/newsfragments/244.bugfix
@@ -1,0 +1,1 @@
+Fix issue where some command errors where not reported as comments on the PR.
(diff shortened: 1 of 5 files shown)


legalsylvain (approved):
Thanks

PR #252, opened 2023-07-30
Add option to set minimum supported branch

sbidoul:
At some point we will not want to continue generating READMEs and publishing wheels for older Odoo versions.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -7,4 +7,4 @@ combine_as_imports=True
 use_parentheses=True
 line_length=88
 known_first_party=oca_github_bot
-known_third_party = aiohttp,appdirs,celery,gidgethub,github3,lxml,odoorpc,pytest,requests,setuptools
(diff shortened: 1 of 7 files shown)


legalsylvain:
Naive question : why ?

sbidoul:
The short answer is that we can't support old versions forever.

In more details
- when I'll deploy the new README.rst generator, it will generate runboat links and runboat does not support Odoo 8 and 9 (it could but nobody contributed nor asked for it)
- the oca/oca-ci image does not support Odoo 8 and 9 (it could but nobody contributed nor asked for it)
- the OCA repo template does not support 8 and 9
- sooner or later we'll stop paying for travis so we won't have CI anymore for 8 and 9, without official alternative since no one tests the repo template with 8 and 9

So I'm going to ask that we declare 8 and 9 officially unsupported by OCA.

sbidoul:
Ah, I forgot to mention that the main branch bot consumes a significant amount of resources on the OCA server, so it will help with that too.

Nevertheless this PR does not change the default behaviour, it just had a configuration option.

pedrobaeza (approved):

PR #256, opened 2023-09-02
Update readme generator and generate README.rst only when fragments changed

sbidoul:


src/oca_github_bot/tasks/main_branch_bot.py
```diff
@@ -33,6 +33,7 @@ def _gen_addons_readme(org, repo, branch, cwd):
     _logger.info("oca-gen-addon-readme in %s/%s@%s", org, repo, branch)
     gen_addon_readme_cmd = [
         "oca-gen-addon-readme",
+        "--if-fragments-changed",
```
legalsylvain, line 36: If i understand correctly, the readme will not be updated, if a PR change only the manifest info (like module name, maintainers, licences, ...)
If yes, It is not desired, dont you think ?

sbidoul: The option name might be misleading but the manifest is included in the digest computation as well as the whole readme fragments directory:

https://github.com/OCA/maintainer-tools/blob/9a5d4cbe17a7fe9921e4157d74e64461019e03f3/tools/gen_addon_readme.py#L516-L520

legalsylvain: Oh. Should be --if-source-changed ?

sbidoul: Alrighty. Done :)

PR #293, opened 2024-06-26
Upgrade to python 3.12

sbidoul:
Python 3.8 is almost EOL, so we upgrade to python 3.12.

We support python 3.12 only, because the dependencies are numerous, and maintaining a set of dependency versions that works with all python versions is too much work. 

Since we deploy from the Dockerfile anyway that should not be an issue.


sbidoul:
@legalsylvain no objection with supporting python 3.12 only?

sbidoul:
If there are no objections I'll go ahead with this soon(ish) to see if this helps me diagnose the new memory consumption issues of the bot...

pedrobaeza:
Please go ahead.

PR #295, opened 2024-07-06
More upgrades

sbidoul:
- upgrade Dockerfile to use python 3.12
- upgrade maintainer tools
- upgrade other deps


Dockerfile
```diff
@@ -24,7 +26,7 @@ RUN set -x \
 # Install a specific version of icon generator, to ensure stability as any tiny change
 # in generated output may create many commits on all addons.
 RUN set -x \
-  && python3 -m venv /ocamt-pinned \
+  && python${PY} -m venv /ocamt-pinned \
   && /ocamt-pinned/bin/pip install --no-cache-dir -U pip wheel
 RUN set -x \
   && /ocamt-pinned/bin/pip install --no-cache-dir -e git+https://github.com/OCA/maintainer-tools@969238e47c07d0c40573acff81d170f63245d738#egg=oca-maintainers-tools \
```
pedrobaeza, line 32: Shouldn't this one be updated as well?

sbidoul: Not now. We'll need to check if something important has changed there first.

PR #299, opened 2024-07-22
Publish to PyPI immediately before pushing to main branch

sbidoul:
We publish to PyPI before merging, because we don't want to merge if PyPI rejects the upload for any reason (such as ownership issues). There is a possibility that the upload succeeds and then the merge fails, but that should be  exceptional, and it is better than the contrary.


legalsylvain (approved):

simahawk (approved):

sbidoul:
This has been deployed earlier this week.

PR #300, opened 2024-07-22
Label PRs with modified addons

### LINKED ISSUE #265: Autolabel PRs with the name of new / modified addons
Got this idea after a conversation with @sebastienbeau 

**Is your feature request related to a problem?**

In OCA it is not easy to get an overview of the PRs that impact addons you are interested in.

**Describe the solution you'd like**

The bot already has logic to know which modules are added/updated by a PR.
When a PR is created or when new commits are pushed to a PR the bot could look at modified addons and label the PR with `{type}: {addon_name}`. Where type would describe what happens (such as `addon new`, `addon upd`, `addon mig`).

sbidoul:
closes #265 

This adds a `addon:{addon_name}` and a `series:{branch_name}` labels. Note: distinguishing by modification type can be added later.

With this it should become much easier to write tools to be notified about modifications to addons of interest.

### DIFF
--- a/src/oca_github_bot/tasks/label_modified_addons.py
+++ b/src/oca_github_bot/tasks/label_modified_addons.py
@@ -1,0 +1,37 @@
+# Copyright (c) ACSONE SA/NV 2024
+# Distributed under the MIT License (http://opensource.org/licenses/MIT).
+
+from .. import github
+from ..config import switchable
+from ..manifest import git_modified_addons
+from ..process import check_call
+from ..queue import task
+from ..version_branch import is_main_branch_bot_branch
+
+
+@task()
+@switchable("label_modified_addons")
+def label_modified_addons(org, repo, pr, dry_run):
+    with github.login() as gh:
+        gh_pr = gh.pull_request(org, repo, pr)
+        target_branch = gh_pr.base.ref
+        pr_branch = f"tmp-pr-{pr}"
+        with github.temporary_clone(org, repo, target_branch) as clone_dir:
+            check_call(
+                ["git", "fetch", "origin", f"pull/{pr}/head:{pr_branch}"],
+                cwd=clone_dir,
+            )
+            check_call(["git", "checkout", pr_branch], cwd=clone_dir)
+            modified_addons, _ = git_modified_addons(clone_dir, target_branch)
+            if not modified_addons:
+                return
+            gh_issue = github.gh_call(gh_pr.issue)
+            new_labels = {
+                f"addon:{modified_addon}" for modified_addon in modified_addons
+            }
+            if is_main_branch_bot_branch(target_branch):
+                new_labels.add(f"series:{target_branch}")
+            new_labels = new_labels - {label.name for label in gh_issue.labels()}
+            if new_labels and not dry_run:
+                for new_label in new_labels:
+                    github.gh_call(gh_issue.add_labels, new_label)
--- a/src/oca_github_bot/tasks/__init__.py
+++ b/src/oca_github_bot/tasks/__init__.py
@@ -3,6 +3,7 @@
 
 from . import (
     heartbeat,
+    label_modified_addons,
     main_branch_bot,
     mention_maintainer,
     migration_issue_bot,
--- a/src/oca_github_bot/version_branch.py
+++ b/src/oca_github_bot/version_branch.py
@@ -33,7 +33,7 @@ def is_main_branch_bot_branch(branch_name):
 
 
 def is_protected_branch(branch_name):
-    if branch_name == "master":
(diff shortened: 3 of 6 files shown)


hhgabelgaard:
@sbidoul Why not use Milestone for the branch series tagging?
Like its done with the migration issues
And manually in som OCA repos

sbidoul:
> Why not use Milestone for the branch series tagging? Like its done with the migration issues And manually in som OCA repos

We could and probably should.

But I always have been thinking that this is a misuse of Milestones which are meant to manage the next release of a project. On small repos like MIS Builder I have sometimes used the milestone to that end.

sbidoul:
Interesting approach. My initial use case was to search by label though so that would be a bit more complicated as we'd need a helper to compute the label with hash from the addon name. But feasible I guess.

legalsylvain:
> Interesting approach. My initial use case was to search by label though so that would be a bit more complicated as we'd need a helper to compute the label with hash from the addon name. But feasible I guess.


- if you want to implement search for a tool / by API, you can easily implement the 5 lines that compute the name with the hash.
- if you want to use it manually, you can do it without knowing the hash, because the search is done in the description field. Step to reproduce : 

-> go to https://github.com/grap/grap-odoo-business/pulls
-> Open the label filter search bar
-> enter the full name of your module (account_invoice_supplierinfo_update_standard_price)
-> the label with the hash is proposed. (the description with the full name is proposed below).

image


As a result : 
- it is possible to search & filter
- The label name will look a bit ugly for 3–5% of OCA modules. (not a big deal). 3-5 is approximative, because I could not find in odoo-community.org the list of all OCA modules. (like in V14, I think that there is a pending refactor regarding the github-connector features) However, I honestly don't think there are many modules that are longer than 43 characters. (``50 - len("addons:")``)

WDYT ?

sbidoul:
> WDYT ?

I think good things :)  Sounds cool. Thanks for picking up this topic!

legalsylvain:
Ready to be merged for me. 
@hhgabelgaard, @manuel-florido, @ValentinVinagre, @sebastienbeau a final review ? 

See #265

src/oca_github_bot/tasks/label_modified_addons.py
```diff
@@ -0,0 +1,56 @@
...
+                github.gh_call(
+                    gh_repo.create_label,
+                    name=label_name,
+                    description=f"Module {modified_addon}",
+                    color=MODULE_LABEL_COLOR.replace("#", ""),
+                )
+            new_labels.add(label_name)
+
+        if is_main_branch_bot_branch(target_branch):
+            new_labels.add(f"series:{target_branch}")
```
legalsylvain, line 47:
```suggestion
            new_labels.add(f"serie:{target_branch}")
```

I use this branch in production since a few days and LGTM.
But I think that the "serie" word should be singular, not plural. Don't you think ?

sbidoul: `series` with s is singular, I think: https://www.merriam-webster.com/grammar/series-singular-plural-grammar-usage

sbidoul:
If no one has objections, I'll deploy this in OCA in the coming days.

src/oca_github_bot/tasks/label_modified_addons.py
legalsylvain: Thanks for reviewing my english ;-) 

Indeed.

manuel-florido:
LGTM 

Non-blocking suggestion: gh_repo.labels() is called inside the loop, so the repository labels are checked once per modified addon. It could be slightly more efficient to fetch the existing label names once before the loop and reuse them.

One question: on synchronize, only adds missing labels, but it does not remove module labels that may have become obsolete if the set of modified addons changed after a push or force-push. Is that the intended behavior?

@sbidoul @legalsylvain @ValentinVinagre

sbidoul:
Good catches. At least the first item would be good to fix because the bot is always close to reaching GitHub API limits.

legalsylvain:
hi @manuel-florido and @sbidoul. I implemented your suggestions and deploy in my ocabot instance.

# Test PR  

https://github.com/grap/grap-odoo-incubator/pull/217

# Result

image

could you review https://github.com/acsone/oca-github-bot/pull/6

thanks !

manuel-florido:
This [MR](https://github.com/acsone/oca-github-bot/pull/6) resolves my previous comments.

LGTM, thanks!

@legalsylvain @sbidoul

sbidoul:
Deployed in OCA and it appears to work. Example: https://github.com/OCA/manufacture/pull/1763

Thanks a lot for driving this to completion, @legalsylvain !

PR #319, opened 2025-05-12
Upgrade maintainer-tools

sbidoul:
Mostly to get the new repo readme renderer from https://github.com/OCA/maintainer-tools/pull/653


pedrobaeza (approved):

pedrobaeza:
Check the conflicts, and thinking twice, is this going to provoke a difference between the README generator in each pre-commit with this one? I have observed that some READMEs are rendered differently when you pass pre-commit, and may be in part due to this. Can't the pre-commit version of each repository be used instead?

sbidoul:
@pedrobaeza the bot manages the repo README.md which is not managed by precommit, so it should be a problem.

sbidoul:
Updated to get the README banner from https://github.com/OCA/maintainer-tools/pull/656

PR #320, opened 2025-08-12
[FIX] Update other release branches

yajo:
These releases are quite old, and modern ones are not taken into
consideration.

I know this is just a sample, but please take this chance to look at the
current bot production deployment, to see if this is misconfigured too.

@moduon MT-10479

### DIFF
--- a/environment.sample
+++ b/environment.sample
@@ -83,4 +83,4 @@ OCABOT_TWINE_REPOSITORIES="[('https://pypi.org/simple','https://upload.pypi.org/
 
 # List of branches the bot will check to verify if user is the maintainer
 # of module(s)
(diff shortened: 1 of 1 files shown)


sbidoul:
In the OCA bot instance container, now (and since we updated it to support 18.0):

```
$ echo $MAINTAINER_CHECK_ODOO_RELEASES
10.0,11.0,12.0,13.0,14.0,15.0,16.0,17.0,18.0
```

yajo:
@sbidoul Do you mean that that was that the configuration already running, or that you changed it to that now?

I assumed it was wrongly configured and that's why we had https://github.com/OCA/timesheet/pull/809#issuecomment-3178865555 and some other similar issues I've seen.

@legalsylvain attended.

sbidoul (approved):

manuel-florido (approved):
LGTM!

PR #327, opened 2026-01-11
[IMP] /ocabot merge: ignore skipped check suites

### LINKED ISSUE #326: /ocabot merge doesn't proceed with skipped workflows
see https://github.com/OCA/OpenUpgrade/pull/5375#issuecomment-3689353804

I suspect we need to change https://github.com/OCA/oca-github-bot/blob/master/src/oca_github_bot/tasks/merge_bot.py#L336, but then https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#list-commit-statuses-for-a-reference doesn't explicitly mention the value 'skipped'

hbrunn:
In OpenUpgrade, we have a workflow that is [skipped](https://github.com/OCA/OpenUpgrade/blob/18.0/.github/workflows/generate-analysis-cron.yml#L14) most of the time because it's super resource consuming and only makes sense to be run after merge on the main branch anyways.

Without this, ocabot treats commits with skipped workflows as [failed](https://github.com/OCA/OpenUpgrade/pull/5443#issuecomment-3733034406), which it shouldn't.

fixes #326

### DIFF
--- a/src/oca_github_bot/tasks/merge_bot.py
+++ b/src/oca_github_bot/tasks/merge_bot.py
@@ -371,6 +371,13 @@ def _get_commit_success(org, repo, pr, gh_commit):
                 f"PR #{pr} of {org}/{repo}"
             )
             success = True
+        elif check_suite.conclusion == "skipped":
+            # skipped
+            _logger.info(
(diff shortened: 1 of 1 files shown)


sbidoul:
Thanks!

PR #360, opened 2026-09-03
Upgrade to python 3.14 and use Ubuntu 26.04 in Dockerfile

sbidoul:


pedrobaeza:
Although this is always welcome, as we have seen lately, the real problem is that GitHub is rate-limiting unauthenticated downloads, so we will need to put in the ocabot server some GitHub credentials to work (and pray that GitHub is not rate-limiting also per account downloads).

sbidoul:
All I know at the moment is that GitHub is having troubles. Moving everything to authenticated git operations is not going to be trivial, so lets not rush on that and leave GitHub some time to fix their issues.

pedrobaeza:
Well, all follows that they are moving to this. Let's wait, but freezing whole OCA these days is an annoyance...

sbidoul:
In the meantime I have deployed this branch. Can you try a merge?

pedrobaeza:
For now it's working:

https://github.com/OCA/helpdesk/pull/1100#issuecomment-5522888810

but if it's rate-limiting, it will appear soon.

pedrobaeza:
Officially confirmed that they are applying limits to unauthenticated users: https://github.com/orgs/community/discussions/206581#discussioncomment-18269083

So although for now they are allowing more rate, it's a matter of time.

sbidoul:
@legalsylvain no objection for this one?

